Issue #51, opened 2024-04-04
java.lang.ClassCastException: android.net.Uri$StringUri cannot be cast to java.lang.String


Qw4z1:
Similar to #48, I get an exception when trying to use `public expect val CustomerInfo.managementURL: String?`. 

This is with: 
Kotlin version 1.9.21
Compose version 1.5.11

See simplified example below:

```
data class ManagementUrl(val url: String)

PurchasesFactory.sharedInstance.awaitCustomerInfo().let { customerInfo ->
    **val url: String? = customerInfo.managementURL** // This is where the crash happens
    ManagementUrl(url)
}
``` 

For whatever reason, adding identical expect/actual implementations with a slightly different name works.

```
expect val CustomerInfo.managementURLString: String?

actual val CustomerInfo.managementURLString: String?
    get() = managementURL?.toString()
```

JayShortway:
Thanks for flagging this! What kind of exception are you getting? Do you have a stacktrace?

Qw4z1:
Ah, my bad. `val url: String? = customerInfo.managementURL **as String?**` crashes on both API 34 and 28.

If I remove `as String?` I get a compile time error instead saying `Type mismatch: inferred type is Uri? but String? was expected`. So it seems like it's looking at the RevenueCat SDK Android implementation of `CustomerInfo.managementUrl`, which is the one returning a `URI?`.

JayShortway:
Thanks, your analysis seems very plausible! I'll have to make sure that the Android implementation returns a String instead of a Uri.

JayShortway:
I managed to reproduce it as part of #56, which will be run on CI to prevent regressions once that is merged. 

I also have an idea what's going wrong. The extension shadows the official SDK implementation, i.e. has the same name. The compiler seems to be picking that one, which returns a `Uri`. It then throws a type error because the field is defined as a `String`. That's also why your workaround of adding identical expect/actual implementations with a slightly different name works. The name is exactly the issue :) 

Will see what the most elegant fix is for this. (And see if there's even any other option than simply renaming the extension.)

Issue #48, opened 2024-03-30
java.lang.ClassCastException: com.revenuecat.purchases.models.GoogleStoreProduct cannot be cast to io.shortway.kobankat.models.StoreProduct


Moozart:
Hello,

When I use map like this in composeApp module 

`availablePackages.map { pack -> pack.product as StoreProduct}`

Throws java.lang.ClassCastException: com.revenuecat.purchases.models.GoogleStoreProduct cannot be cast to io.shortway.kobankat.models.StoreProduct

If don't add type casting, throws build time  error

Type mismatch: inferred type is com.revenuecat.purchases.models.StoreProduct but io.shortway.kobankat.models.StoreProduct was expected

Do you have any idea or advice.
Thanks to you for good libray

JayShortway:
Hi, thanks for flagging this!

If you don't add type casting, which target are you building that throws the build error? 
What's the reason for the cast? To attempt to fix the build error, or is there an additional reason? 

I'm currently on holiday so I can't try to reproduce it myself unfortunately.

Moozart:
Target Android has this error. IOS not. I just want to copy products in cacheStoreProductsList: List<StoreProduct> to purchase it via productId finding from cacheStoreProductList. Because I pass only Id to uiState. 

I changed my strategy after that crash. I used Package to purchase via identifier. There is no issue with Package

Qw4z1:
I'm also seeing `Type mismatch: inferred type is com.revenuecat.purchases.models.StoreProduct but io.shortway.kobankat.models.StoreProduct was expected` at compile time, even though I'm importing `io.shortway.kobankat.models.StoreProduct` and just call `offering.package.product` to get the StoreProduct.

Could it be [this line](https://github.com/JayShortway/kobankat/blob/main/core/src/androidMain/kotlin/io/shortway/kobankat/models/StoreProduct.android.kt#L10) that is confusing things?

JayShortway:
@Moozart Good to hear you found a workaround!
@Qw4z1 I think you might be right that that line is the culprit! Thanks for pinpointing it.

For some reason StoreProduct is not a type alias, while most other classes are. I'm not sure what the reason for that was anymore.

Since I'm currently on holiday, I'll be able to look at this in about 2 weeks or so.

Qw4z1:
@JayShortway No stress! You've done plenty already! =)

I might take a stab at it if I get time before you're back. Not my area of expertise, though, so no promises. 😅

[...]

PR #80, opened 2024-06-10
Changes package to com.revenuecat

JayShortway:
Lots of files changed, due to the nature of the change. 🙁 

## Changed
- Changes the package name from `io.shortway.kobankat` to `com.revenuecat.purchases.kmp`. (The `.kmp` suffix is added to avoid conflicts with the Android SDK in the `androidMain` source set.)
- Updates names of run configurations, removing the "KobanKat" name. Also added 2 more run configurations: apiDump and apiCheck.
- Changes the application/bundle ID of the sample app to that of PurchaseTester. 
- Updates `renovate.json` so it runs for the updated repository name (`purchases-kmp`). Not sure if we're going to keep it though.   

## Unchanged
- Did not touch any of the publishing logic yet, so that still references KobanKat. Will do that later.


core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/EntitlementInfos.android.kt
```diff
@@ -9,4 +9,4 @@ public actual typealias EntitlementInfos = RcEntitlementInfos
 public actual val EntitlementInfos.all: Map<String, EntitlementInfo>
     get() = all
 public actual val EntitlementInfos.verification: VerificationResult
-    get() = verification
\ No newline at end of file
+    get() = verification
```
tonidero, line 12: Hmm I'm wondering, what about the other properties for `EntitlementInfos` (like `active`, or `get`)? Maybe we still need to add them here but I wanted to check

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/models/Price.ios.kt
```diff
@@ -25,7 +23,7 @@ public actual data class Price(
 
 internal fun RCStoreProduct.toPrice(): Price =
     localizedPriceString().let { localizedPrice ->
-        Price(
+        com.revenuecat.purchases.kmp.models.Price(
```
tonidero, line 26: Hmm is it name colliding with something else? Being in the same file, I would assume it would use the `Price` defined here automatically, but not sure...

datetime/src/commonMain/kotlin/com/revenuecat/purchases/kmp/datetime/CustomerInfo.kt
```diff
@@ -1,14 +1,14 @@
-package io.shortway.kobankat.datetime
+package com.revenuecat.purchases.kmp.datetime
```
tonidero, line 1: Any reason to have these in a separate module? We haven't been doing much modularization in `purchases-android` since we thought we didn't need it at least right now but would be good to know the reasoning.

paywalls/src/androidMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/Paywall.kt
```diff
@@ -1,4 +1,4 @@
-package io.shortway.kobankat.ui.revenuecatui
+package com.revenuecat.purchases.kmp.ui.revenuecatui
```
tonidero, line 1: We should rename this module from `paywalls` to `revenuecatui` or something like that, since we might eventually add more UI other than paywalls here?

tonidero (commented):
Left some questions, most of them unrelated to the changes here, but just in case we can discuss them to address in future PRs.

apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/PeriodAPI.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.revenuecat.purchases.kmp.models.unit
+import com.revenuecat.purchases.kmp.models.value
+import com.revenuecat.purchases.kmp.models.valueInMonths
+
+@Suppress("unused", "UNUSED_VARIABLE")
+private class PeriodAPI {
+    fun check(period: Period) {
+        with(period) {
+            val value: Int = value
+            val unit: PeriodUnit = unit
```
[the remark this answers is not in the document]
JayShortway: Definitely! The current API tests are just the most important ones. Added a Linear ticket: [SDK-3480](https://linear.app/revenuecat/issue/SDK-3480).

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/App.kt
```diff
@@ -1,4 +1,4 @@
-package io.shortway.kobankat.sample
+package com.revenuecat.purchases.kmp.sample
```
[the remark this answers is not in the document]
JayShortway: Yea good points. The current `composeApp`/`iosApp` is indeed more for us to develop the library. I was planning on adding examples and changing folder structure in [SDK-3307](https://linear.app/revenuecat/issue/SDK-3307).

core/api/core.api
```diff
@@ -1,4 +1,4 @@
-public final class io/shortway/kobankat/CustomerInfo_androidKt {
+public final class com/revenuecat/purchases/kmp/CustomerInfo_androidKt {
```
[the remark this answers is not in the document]
JayShortway: I agree that in theory we wouldn't need manual API tests. However, I added them because of these issues which they reproduce: https://github.com/RevenueCat/purchases-kmp/issues/48, https://github.com/RevenueCat/purchases-kmp/issues/51. This was only found by calling a method/field and compiling that call. 

The underlying issue here is an extension shadowing a member, but changing its type. I'm planning to catch those issues with a Detekt rule [SDK-3464](https://linear.app/revenuecat/issue/SDK-3464). So if we have that rule, then we might consider removing the manual API tests.

datetime/src/commonMain/kotlin/com/revenuecat/purchases/kmp/datetime/CustomerInfo.kt
JayShortway: They add a dependency on `kotlinx-datetime` which I didn't want to impose on people if they don't use it already. Although it is probably the most popular datetime library, they might also be using [Klock](https://docs.korge.org/time/) or [Island Time](https://islandtime.io/), for instance.

paywalls/src/androidMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/Paywall.kt
JayShortway: `revenuecatui` can work yes! I named it `ui` at first, like in purchases-android, but got various conflicts. Don't exactly remember what it was, but maybe Xcode didn't like it. 

Of course the Gradle module name and published artifact ID don't have to be the same, so we can still publish this module as artifact `:ui`, and change the Gradle module name when we build more UI stuff?

apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/StoreProductAPI.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.revenuecat.purchases.kmp.models.price
+import com.revenuecat.purchases.kmp.models.purchasingData
+import com.revenuecat.purchases.kmp.models.subscriptionOptions
+import com.revenuecat.purchases.kmp.models.title
+import com.revenuecat.purchases.kmp.models.type
+
+@Suppress("unused", "UNUSED_VARIABLE", "LongMethod")
+private class StoreProductAPI {
+    fun check(product: StoreProduct) {
+        val locale = Locale.Default
```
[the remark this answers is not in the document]
JayShortway: No, you're right! This is a leftover. Deleted: https://github.com/RevenueCat/purchases-kmp/pull/80/commits/253aef3fb298b9568b7d85387194a8b00435fc25.

paywalls/src/androidMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/Paywall.kt
tonidero: Yeah I think that's ok for now 👍

tonidero (approved):
Thanks for answering my questions! This looks good to merge :)

datetime/src/commonMain/kotlin/com/revenuecat/purchases/kmp/datetime/CustomerInfo.kt
tonidero: I do wonder if we want to be more opinionated with these, since it would make for a much easier structure... But I'm ok keeping it separate for now. At the end of the day, this is #customer-obsession

core/api/core.api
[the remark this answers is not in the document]
tonidero: Nice! Thanks for the explanation!

datetime/src/commonMain/kotlin/com/revenuecat/purchases/kmp/datetime/CustomerInfo.kt
JayShortway: Haha yea true. It's something I want to discuss with the team before the release!

paywalls/src/androidMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/Paywall.kt
JayShortway: So I tried to name it `ui` again to check if this was a fluke, but it really doesn't work haha. I'm getting these errors related to Compose Multiplatform on iOS:

>Cannot access class 'androidx.compose.ui.Modifier'. Check your module classpath for missing or conflicting dependencies. 
  
Since these are part of the `compose.ui` package, maybe it conflicts on the `ui` part there? Pretty weird.

PR #83, opened 2024-06-12
Adds snapshot publishing

JayShortway:
## Changes
- Adds a CircleCI config with a `deploy-snapshot` job that runs for all pushes to `main`. 
- Removes the NmcpPlugin, which I had to use in KobanKat. I had created my Maven Central account after a certain date which meant I could only publish via the [Central Portal](https://central.sonatype.org/register/central-portal/), and `com.vanniktech.maven.publish` did not have support for that at that time. (See https://github.com/vanniktech/gradle-maven-publish-plugin/issues/722.)
- Adjusts the coordinates and POM file to contain RevenueCat info instead of my personal info.
- Updates the klib abi dump, because the group changed (to `com.revenuecat.purchases`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,31 @@
+# Use the latest 2.1 version of CircleCI pipeline process engine.
+# See: https://circleci.com/docs/configuration-reference
+version: 2.1
+
+# Define a job to be invoked later in a workflow.
+# See: https://circleci.com/docs/jobs-steps/#jobs-overview & https://circleci.com/docs/configuration-reference/#jobs
+jobs:
+  say-hello:
+    # Specify the execution environment. You can specify an image from Docker Hub or use one of our convenience images from CircleCI's Developer Hub.
+    # See: https://circleci.com/docs/executor-intro/ & https://circleci.com/docs/configuration-reference/#executor-job
+    docker:
+      # Specify the version you desire here
+      # See: https://circleci.com/developer/images/image/cimg/base
+      - image: cimg/base:current
+
+    # Add steps to the job
+    # See: https://circleci.com/docs/jobs-steps/#steps-overview & https://circleci.com/docs/configuration-reference/#steps
+    steps:
+      # Checkout the code as the first step.
+      - checkout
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,78 @@
...
+          command: |
+            set -euo pipefail
+            file_path="gradle/libs.versions.toml"
+            version=$(grep 'revenuecat-kmp = ' "$file_path" | awk -F ' = ' '{print $2}' | tr -d '"')
+            if [[ ! "$version" =~ -SNAPSHOT$ ]]; then
+              echo "$version is not a SNAPSHOT version. Exiting..."
+              exit 1
+            else
+              echo "$version is a SNAPSHOT version. Proceeding..."
+            fi
```
vegaro, line 36: not needed for now, but this should be moved to fastlane because it's easier to maintain and reuse. It's easier to use ruby than bash

vegaro (approved):

.circleci/config.yml
```diff
@@ -0,0 +1,78 @@
...
+
+executors:
+  xcode15:
+    macos:
+      xcode: 15.4.0
+    resource_class: macos.m1.medium.gen1
+    environment:
+      HOMEBREW_NO_AUTO_UPDATE: 1
```
tonidero, line 44: Just curious why is this needed? Maybe we can put a comment explaining why.

tonidero (approved):

.circleci/config.yml
JayShortway: Homebrew will check for updates every time you call `brew`. This disables that behavior so we don't have to wait for it. Everything we need is freshly installed, so that's already up to date. 
It's also a recommendation from CircleCI: https://circleci.com/docs/using-macos/#optimizing-homebrew

Added a comment with some details!

.circleci/config.yml
JayShortway: Yes agreed! Although for now I don't even know how permanent this behavior is going to be, but will definitely move it to fastlane if it stays.

PR #86, opened 2024-06-13
Corrects the artifactId of the :paywalls module.

JayShortway:
As the title states.


build.gradle.kts
```diff
@@ -29,9 +29,17 @@ allprojects {
             publishToMavenCentral(SonatypeHost.DEFAULT, automaticRelease = true)
             signAllPublications()
 
+            // We override the artifact ID of :paywalls for consistency with the other SDKs. We
+            // could not name our Gradle module :ui, because this somehow conflicts with compose.ui
```
tonidero, line 33: Could we rename the module to `revenuecatui` even if it doesn't match the `purchases-android` structure? I think that might be a better temporal name than `paywalls`. We can still do this to have the artifact name we want though

vegaro (approved):

build.gradle.kts
JayShortway: Sure, done!

tonidero (approved):

PR #90, opened 2024-06-14
Adds Fastlane

JayShortway:
This PR just adds the Fastlane scaffolding. No lanes are added whatsoever. 

Let me know if I overlooked something!


vegaro (approved):
Looks good! One thing that you could add is an empty `.env.SAMPLE` like in the other repos https://github.com/RevenueCat/purchases-android/blob/main/fastlane/.env.SAMPLE, but that can also be added as soon as you start using env variables. Just mentioning in case you missed that exists 😄

PR #99, opened 2024-06-17
Release train

JayShortway:
Adds the required CircleCI config and Fastlane lanes to be able to couple KMP onto the release train. I tested this as far as I could, e.g. it opens the release PR and generates the changelog correctly. I'm thinking the real test comes when we release it for the first time. 
  
I did not yet set up the scheduled pipeline on CircleCI. Will do that when the SDK is fully ready.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,7 @@ commands:
       - run:
           name: Install Android SDK platform tools
           command: sdkmanager "platform-tools"
+
   require-snapshot-version:
     description: Check that the current version is a SNAPSHOT version
     steps:
@@ -78,6 +79,7 @@ jobs:
             version:<< pipeline.parameters.version >> \
             open_pr:true \
             automatic_release:<< pipeline.parameters.automatic >>
+
   publish:
     description: "Publishes a release, either SNAPSHOT or final."
     executor: xcode15
@@ -110,6 +112,7 @@ jobs:
                 command: ./gradlew publishAllPublicationsToMavenCentralRepository
       - android/save-gradle-cache
       - android/save-build-cache
+
   create-github-release:
     description: "Creates a GitHub release for the current version"
     executor: ruby3
@@ -138,7 +141,6 @@ workflows:
             - github-bot-public
 
   on-main-branch:
-    # On all pushes to main, except when it's a scheduled pipeline.
     when:
       and:
         - equal: [ main, << pipeline.git.branch >> ]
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -93,11 +101,26 @@ jobs:
...
+      # TODO Remove this condition to allow releasing non-SNAPSHOT versions. 
+      - when:
+          condition: << parameters.snapshot >>
+          steps:
+            - run:
+                name: Publish to Maven Central
+                command: ./gradlew publishAllPublicationsToMavenCentralRepository
       - android/save-gradle-cache
       - android/save-build-cache
+  create-github-release:
```
tonidero, line 113: Nitpick but I would leave an empty line between jobs. I think that's how we've been doing it?

.circleci/config.yml
```diff
@@ -114,13 +137,52 @@ workflows:
...
+        # The release-train can be manually triggered by passing the 
+        # "action" parameter with value "bump".
+        - equal: [ bump, << pipeline.parameters.action >> ]
+    jobs:
+      - revenuecat/automatic-bump
+
+  on-release-branch:
+    when:
+      and:
+        - matches: { pattern: "^release/.+$", value: << pipeline.git.branch >> }
```
tonidero, line 167: Hmm interesting... We've usually had `aliases` for these...I guess this applies to the whole workflow, which might be nicer than having to copy the alias on each job, so maybe this is ok.

I have wondered before if it would be possible to move those aliases to our CircleCI orb but haven't taken a look. Not sure if it would be possible with this... In any case, that's something we can do later.

.circleci/config.yml
```diff
@@ -93,11 +101,26 @@ jobs:
             echo 'export ORG_GRADLE_PROJECT_mavenCentralPassword="$SONATYPE_NEXUS_TOKEN_PASSWORD"' >> "$BASH_ENV"
             echo 'export ORG_GRADLE_PROJECT_signingInMemoryKey="$SIGNING_GPG_IN_MEMORY"' >> "$BASH_ENV"
             echo 'export ORG_GRADLE_PROJECT_signingInMemoryKeyPassword="$GPG_SIGNING_KEY_PW_NEW"' >> "$BASH_ENV"
-      - run:
-          name: Publish snapshot to Maven Central
-          command: ./gradlew publishAllPublicationsToMavenCentralRepository
+      # TODO Remove this condition to allow releasing non-SNAPSHOT versions.
```
tonidero, line 104: Hmm I'm wondering if we should just remove this and just be mindful of creating tags in the format X.Y.Z? Seems unlikely to be accidental.

tonidero (approved):
LGTM!

.circleci/config.yml
JayShortway: Yea we could, but this is more so I can sleep soundly at night 😛 Jokes aside, I also wanna refactor this a little bit in a follow-up PR based on a previous [comment](https://github.com/RevenueCat/purchases-kmp/pull/83#discussion_r1637869434) by César, so this is very temporary anyway.

.circleci/config.yml
JayShortway: Ah good catch, done!

.circleci/config.yml
JayShortway: Yea I was wondering what you'd think of this 🙂 I personally think this is more readable than having `when` clauses (through aliases) on each job. It's easier to follow the logic this way. In any case I added some "header docs" to each workflow. 

Looking quickly at the CircleCI docs, I don't think it's possible for an orb to provide aliases. GitLab CI/CD has the concept of [includes](https://docs.gitlab.com/ee/ci/yaml/includes.html), which would allow us to define aliases in a separate YAML file in a central git repo, and then include that in all our other repos. It [doesn't seem like](https://support.circleci.com/hc/en-us/articles/360056463852-Can-I-split-a-config-into-multiple-files) CircleCI has anything like that though.

PR #100, opened 2024-06-18
Moves publishing logic from the CircleCI config to Fastlane.

JayShortway:
As the title says.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -105,6 +105,26 @@ lane :tag_current_branch do |options|
   push_git_tags(tag: version_number)
 end
 
+desc "Publish the SDK if the current version is a SNAPSHOT version"
+lane :publish_if_snapshot do |options|
+  version = current_version_number
+  if version.end_with?("-SNAPSHOT")
+    publish
+  else
+    UI.user_error!("The current version '#{version}' is not a SNAPSHOT version.")
+  end
+end
+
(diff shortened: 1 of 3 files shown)


fastlane/Fastfile
```diff
@@ -105,6 +105,26 @@ lane :tag_current_branch do |options|
   push_git_tags(tag: version_number)
 end
 
+desc "Publish the SDK if the current version is a SNAPSHOT version"
+lane :publish_if_snapshot do |options|
```
tonidero, line 109: Hmm not sure if it would be better to unify these 2 lanes and just pass a parameter, which can be the one set in CircleCI, `parameters.snapshot`... but FFTI. That's just a preference :P

tonidero (approved):
Looks great! Left a comment, but FFTI

fastlane/Fastfile
JayShortway: That's a nice idea! But I'm not sure how to express the same behavior with a parameter. Right now it fails if you run `publish_if_snapshot` while the currently specified version is not a `SNAPSHOT` (and vice versa), which is intended. (That's why there's an `if` in the name.) With a parameter it would look something like: 

```shell
publish snapshot:true
```

which reads as if it will publish a snapshot version regardless of the currently specified version. 

I'm inclined to keep this idea "in the back pocket", and use it if this purchasing logic changes. Let me know if you feel otherwise!

PR #102, opened 2024-06-18
Adds prepare-next-snapshot-version CI job

JayShortway:
As the title says.


fastlane/Fastfile
```diff
@@ -14,6 +14,7 @@ REPO_NAME = "purchases-kmp"
 PATH_CHANGELOG_LATEST = './CHANGELOG-LATEST.md'
 PATH_CHANGELOG = './CHANGELOG.md'
 PATH_VERSIONS = './VERSIONS.md'
+FILE_WITH_SDK_VERSION = { './gradle/libs.versions.toml' => ["revenuecat-kmp = \"{x}\""] }
```
tonidero, line 17: Nitpick, but since this is a map, I would make this plural: `FILES_WITH_SDK_VERSION`

tonidero (approved):
LGTM!

fastlane/Fastfile
JayShortway: But I was so happy it's just 1 file 😛 I hope we can keep that up as long as possible haha.

JayShortway: Updated!

PR #103, opened 2024-06-19
Remove PurchasesFactory

JayShortway:
## Background & motivation
`Purchases` used to be a `typealias` on each platform, which prevented us from accessing its companion object to call static functions (such as `configure()`). This meant it was necessary to have a separate object to hold these functions, which was named `PurchasesFactory`. This is however very inconsistent with the other SDKs. 

This PR refactors `Purchases` to be a thin wrapper around the platform implementations instead. This way it can also expose the static functions in its own companion object, making the public API consistent with the other SDKs. 

## A note on PR size
The most important change is in `Purchases.kt`, and perhaps `PurchasesCommonAPI.kt`. Lots of changes are in the `.api` files. Another big chunk is updating references in documentation.


core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -1,288 +1,370 @@
-@file:Suppress("EXTENSION_SHADOWED_BY_MEMBER")
+@file:Suppress("ACTUAL_ANNOTATIONS_NOT_MATCH_EXPECT")
```
tonidero, line 1: Hmm curious which members aren't matching the expect? Just wondering since this seems like a useful warning to have. If it is needed, might be good adding a comment explaining the need.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -1,288 +1,370 @@
...
+            set(value) = IosPurchases.setSimulatesAskToBuyInSandbox(value)
+
+        public actual var forceUniversalAppStore: Boolean
+            get() = IosPurchases.forceUniversalAppStore()
+            set(value) = IosPurchases.setForceUniversalAppStore(value)
+
+        public actual fun configure(
+            configuration: PurchasesConfiguration
+        ): Purchases {
+            // TODO Enable this in SDK-3301 checkCommonVersion()
```
tonidero, line 56: Hmm is this comment outdated? Not sure what it refers to (the `checkCommonVersion` method is below)

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -1,309 +1,373 @@
...
-        finishTransactions = value
+import java.net.URL
+import com.revenuecat.purchases.DangerousSettings as AndroidDangerousSettings
+import com.revenuecat.purchases.Purchases as AndroidPurchases
+import com.revenuecat.purchases.hybridcommon.configure as commonConfigure
+
+public actual class Purchases private constructor(private val androidPurchases: AndroidPurchases) {
+    public actual companion object {
+
+        private var _sharedInstance: Purchases? = null
```
tonidero, line 35: Hmm I wonder if we should keep the instance here or just rely on the native layer... It would make the instance methods just call a static method, similar to how you are doing the `logHandler` I guess? Not sure, I think this is ok for now though!

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -1,288 +1,370 @@
...
-public actual var Purchases.finishTransactions: Boolean
-    get() = finishTransactions()
-    set(value) {
-        setFinishTransactions(value)
+import com.revenuecat.purchases.kmp.strings.ConfigureStrings
+import platform.Foundation.NSURL
+import cocoapods.PurchasesHybridCommon.RCDangerousSettings as IosDangerousSettings
+import cocoapods.PurchasesHybridCommon.RCPurchases as IosPurchases
+
+public actual class Purchases(private val iosPurchases: IosPurchases) {
```
tonidero, line 24: Do we need to make the constructor `private` as well? So it can't be instantiated like that... I see we do it in the android version, so not sure.

tonidero (approved):
It's great to get a more consistent API :)

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
JayShortway: It refers to these PRs 😄 
- https://github.com/RevenueCat/purchases-hybrid-common/pull/838
- https://github.com/RevenueCat/purchases-kmp/pull/104

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
JayShortway: Hmm, but we would be creating new `Purchases` wrappers then every time the user calls `sharedInstance`? Or did you have something else in mind?  
  
This actually prompted me to change `logHandler` to also keep a reference. 😅 The reason being that the user will expect the `logHandler` field to hold exactly what they set it to. If we don't keep a reference, we are converting to and from the platform type on `set` and `get`. That means if they set it to an instance of `MyAwesomeLogHandler`, they expect to be able to get the same `MyAwesomeLogHandler` instance by reading `Purchases.logHandler`. If we are converting to and from the platform type, they will just receive an anonymous object implementing `LogHandler`.

Let me know if you disagree!

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
JayShortway: Yes! Good catch! https://github.com/RevenueCat/purchases-kmp/pull/103/commits/be80594f5cbe89947109a46c111ea5a3f03344d7

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
tonidero: Hmm right, if the instance is not the same it could lead to issues... Yeah I think it makes sense, let's keep it this way then!

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
JayShortway: I removed the warning and with that the need for this suppression!
 
It suppressed a warning about a missing `@JvmStatic` annotation (which was present in the commonMain source set):
>Annotation `@JvmStatic` is missing on actual declaration.  
  
However that annotation cannot be set in the iosMain source set, so it was impossible to satisfy the linter, basically. 

Removing the annotations in the commonMain source set and only keeping them in androidMain works just as well, but without warnings. The only thing I could not get to work is `@JvmOverloads` on `canMakePayments()`. We can try that again later imo.

PR #105, opened 2024-06-20
Uses PurchasesAreCompletedBy

JayShortway:
Uses the new `PurchasesAreCompletedBy` api everywhere, and removes all mentions of "observer mode" and "finish transactions". 

Unfortunately PHC does not use purchases-android 8.0.0 (purchases-ios 5.0.0), so I had to call the now-deprecated API underneath. (And I couldn't type alias the enum on Android.) These are just implementation details though.


core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -67,10 +69,13 @@ public actual class Purchases private constructor(private val androidPurchases:
...
                 // Using the common configure() call allows us to pass PlatformInfo.
                 commonConfigure(
-                    AndroidProvider.requireApplication(),
+                    context = AndroidProvider.requireApplication(),
                     apiKey = apiKey,
                     appUserID = appUserId,
-                    observerMode = observerMode,
+                    observerMode = when (configuration.purchasesAreCompletedBy) {
+                        REVENUECAT -> false
+                        MY_APP -> true
```
tonidero, line 77: We could potentially have this enum as an internal property of the enum, so we don't duplicate the logic for iOS and android. Having said that, that logic would disappear once we update to latest iOS/android, so NABD.

tonidero (approved):

core/api/core.api
```diff
@@ -206,13 +206,21 @@ public final class com/revenuecat/purchases/kmp/Purchases$Companion {
...
+}
+
 public final class com/revenuecat/purchases/kmp/PurchasesConfiguration {
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;ZLjava/lang/String;ZLcom/revenuecat/purchases/Store;ZLcom/revenuecat/purchases/kmp/DangerousSettings;Lcom/revenuecat/purchases/kmp/EntitlementVerificationMode;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Lcom/revenuecat/purchases/kmp/PurchasesAreCompletedBy;Ljava/lang/String;ZLcom/revenuecat/purchases/Store;ZLcom/revenuecat/purchases/kmp/DangerousSettings;Lcom/revenuecat/purchases/kmp/EntitlementVerificationMode;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun getApiKey ()Ljava/lang/String;
 	public final fun getAppUserId ()Ljava/lang/String;
 	public final fun getDangerousSettings ()Lcom/revenuecat/purchases/kmp/DangerousSettings;
 	public final fun getDiagnosticsEnabled ()Z
-	public final fun getObserverMode ()Z
```
tonidero, line 215: I do wonder if we would want to deprecate these instead of removing.... But not sure

JayShortway: I'm not sure, but wouldn't it be unusual to release a library with already-deprecated things? It would make it "on par" with the platform SDKs, but I don't know how many developers will be copy-pasting their Android implementation code to KMP, for instance. (In that scenario it would be beneficial to have as much parity with the platform SDKs as possible.)

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
JayShortway: I like this! (We have the same in Android.) But I went to implement it and got: 
>`when' with expect enum as subject can not be exhaustive without else branch  
  
It can be exhaustive if we do it in the platform source sets, but then we're duplicating again 😅 

So I'd say it's prolly not worth it. Also serves as a good visual indicator that we need to remove this once PHC depends on the latest iOS/Android. 😛

core/api/core.api
tonidero: Well, at the end of the day I think is to try to not break developers that might already be using this... I think it's fair game to do breaking changes in a 0.X version, but the easier we make it for devs the better. So yeah, I think it's not worth it, please ignore me :P

JayShortway: Good point! Unfortunately this is not the only breaking change. I'm going to write a migration guide, and will shout out on Twitter that I'll help anyone migrating 😄

PR #107, opened 2024-06-21
Fixes snapshot publishing

JayShortway:
As the title says. I noticed snapshots weren't publishing. Not entirely sure what broke it, maybe https://github.com/RevenueCat/purchases-kmp/pull/100?

This should fix it. A successful publish job can be seen [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-kmp/140/workflows/91a98391-17e3-40e1-9adb-997c2dbe2707/jobs/176).


vegaro (approved):

Gemfile
```diff
@@ -1,7 +1,9 @@
 source "https://rubygems.org"
 
 gem 'danger'
+gem 'rb-readline'
```
JayShortway, line 4: This was added as a solution to this issue I ran into: https://github.com/fastlane/fastlane/issues/21794.

PR #108, opened 2024-06-24
Fix iOS StoreProduct description & price

### LINKED ISSUE #65: Bug: Incorrect StoreProduct.description
On iOS only, it looks like the `StoreProduct.description` api can sometimes/always return an incorrect string in the form of 
> <RCStoreProduct: some-hex-number>  
  
Reported via Twitter DM.

JayShortway:
## Fixes
- A `ClassCastException` in `Price.ios.kt`.
- The description of a `StoreProduct`. Closes #65.

`StoreProduct.description` was not working on iOS. This is caused by the name of the field. That shadows [the built-in ObjC member `description`](https://developer.apple.com/documentation/objectivec/nsobjectprotocol/1418746-description), which prints a textual representation of the object. That's also the behavior that was seen in the linked issue. The fix is to rename our field (to `localizedDescription`, like in purchases-ios).


apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/StoreProductAPI.kt
```diff
@@ -30,7 +30,7 @@ private class StoreProductAPI {
             val type: ProductType = type
             val price: Price = price
             val title: String = title
-            val description: String? = description
+            val description: String? = localizedDescription
```
tonidero, line 33: Hmm I'm still wondering whether it's worth deprecating `description` and just remove it once we release 1.0, so people using the SDK don't have to deal with these breaking changes right now...

core/api/core.klib.api
```diff
@@ -570,6 +568,8 @@ final val com.revenuecat.purchases.kmp.models/isBasePlan // com.revenuecat.purch
     final fun (com.revenuecat.purchases.kmp.models/SubscriptionOption).<get-isBasePlan>(): kotlin/Boolean // com.revenuecat.purchases.kmp.models/isBasePlan.<get-isBasePlan>|<get-isBasePlan>@com.revenuecat.purchases.kmp.models.SubscriptionOption(){}[0]
 final val com.revenuecat.purchases.kmp.models/isPrepaid // com.revenuecat.purchases.kmp.models/isPrepaid|@com.revenuecat.purchases.kmp.models.SubscriptionOption{}isPrepaid[0]
     final fun (com.revenuecat.purchases.kmp.models/SubscriptionOption).<get-isPrepaid>(): kotlin/Boolean // com.revenuecat.purchases.kmp.models/isPrepaid.<get-isPrepaid>|<get-isPrepaid>@com.revenuecat.purchases.kmp.models.SubscriptionOption(){}[0]
+final val com.revenuecat.purchases.kmp.models/localizedDescription // com.revenuecat.purchases.kmp.models/localizedDescription|@cocoapods.PurchasesHybridCommon.RCStoreProduct{}localizedDescription[0]
+    final fun (cocoapods.PurchasesHybridCommon/RCStoreProduct).<get-localizedDescription>(): kotlin/String? // com.revenuecat.purchases.kmp.models/localizedDescription.<get-localizedDescription>|<get-localizedDescription>@cocoapods.PurchasesHybridCommon.RCStoreProduct(){}[0]
```
tonidero, line 572: I was just noticing this but are we exposing the PHC API as well? We should try to avoid that... I see we are using `api` in android: https://github.com/RevenueCat/purchases-kmp/blob/main/core/build.gradle.kts#L15. Is this needed? As for iOS, I'm not sure if there is a way to avoid leaking the API... (This is not a blocker for this PR, but I think we should address before initial release)

tonidero (approved):
Just a couple comments... I understand we are in version 0.x, but I feel we should be careful with breaking changes all the same. Other than that looks good!

apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/StoreProductAPI.kt
tonidero: 👍 Sounds good! We can keep it like this then

PR #112, opened 2024-06-25
Consolidates Compose versions

JayShortway:
Only uses Compose Multiplatform dependencies (or, to be more precise, only uses dependencies specified by the Compose Multiplatform plugin). This means we don't need to specify 2 Compose versions (1 for Multiplatform and 1 for "Jetpack").


revenuecatui/build.gradle.kts
```diff
@@ -13,7 +13,7 @@ kotlin {
             implementation(compose.ui)
         }
         androidMain.dependencies {
-            implementation(libs.androidx.compose.ui.tooling.preview)
+            implementation(compose.preview)
```
tonidero, line 16: Just checking but this is not a KMP dependency we can add in `commonMain` right? Same for the `uiTooling` below

tonidero (approved):
LGTM! Just a question

revenuecatui/build.gradle.kts
JayShortway: I thought it wasn't possible, but [it is](https://blog.jetbrains.com/kotlin/2024/02/compose-multiplatform-1-6-0-release/#preview-annotation-for-fleet). https://github.com/RevenueCat/purchases-kmp/pull/112/commits/4dae9d8bcaacaf3f912739d620cfad37998749fb

It only works in Fleet for now, but having previews only in the Android source set is not great either. (You'd have to write your Composable in the common source set, your preview in the Android source set, and go back and forth each time.) 

We can always change back if we decide that that's the lesser of 2 evils.

JayShortway: Thanks for raising!

PR #116, opened 2024-06-26
Adds KobanKat migration guide

JayShortway:
As the title says.

### DIFF
--- a/migrations/KobanKat-MIGRATION.md
+++ b/migrations/KobanKat-MIGRATION.md
@@ -1,0 +1,113 @@
+# KobanKat Migration Guide
+The RevenueCat SDK for Kotlin Multiplatform started out as an independent project named _KobanKat_, built by [@JayShortway](https://github.com/JayShortway). Follow this migration guide if you're currently using KobanKat.
+
+## Maven coordinates
+The Maven coordinates of the SDK have changed. 
+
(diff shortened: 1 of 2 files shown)


migrations/KobanKat-MIGRATION.md
```diff
@@ -0,0 +1,113 @@
+# KobanKat Migration Guide
+The RevenueCat SDK for Kotlin Multiplatform started out as an independent project named _KobanKat_, built by [@JayShortway](https://github.com/JayShortway). Follow this migration guide if you're currently using KobanKat.
+
+## Maven coordinates
+The Maven coordinates of the SDK have changed.
```
tonidero, line 5: I think this is more detailed, but I would maybe add an example just in case, something like:
"For example, if you were using `io.shortway.kobankat:kobankat-core:+`, you need to move to `com.revenuecat.purchases:purchases-kmp-core:+`. "

tonidero (approved):
Looks good!

migrations/KobanKat-MIGRATION.md
JayShortway: Yup that's helpful! Added https://github.com/RevenueCat/purchases-kmp/pull/116/commits/4d94e8a83020996ddaa92f6cd42403d8709381ae

PR #119, opened 2024-06-26
Fixes the release pipeline

JayShortway:
### Adds release tag filters to jobs as well.
As the title says. CircleCI is weird. This bug seems to have existed for a long time where workflows with a tag filter need to have the same filter on jobs as well. 

One of many: https://discuss.circleci.com/t/running-a-workflow-after-tagging-commit/49581. 

Without this change, we get a "[No workflow](https://app.circleci.com/pipelines/github/RevenueCat/purchases-kmp/171)" state from CircleCI.  
  
After this change, [the workflow runs](https://app.circleci.com/pipelines/github/RevenueCat/purchases-kmp/177). 

### Disables configuration cache for the publish task
This is [not supported](https://app.circleci.com/pipelines/github/RevenueCat/purchases-kmp/175/workflows/2c1d41a1-16ad-4b98-af02-6047f9f36763/jobs/251/parallel-runs/0/steps/0-113).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,19 @@
+aliases:
+  # Uses the same regex as release-tags-job
+  release-tags-workflow: &release-tags-workflow
+    when:
+      and:
+        - matches: { pattern: "^[0-9]+\\.[0-9]+\\.[0-9]+(?!.*-SNAPSHOT).*$", value: << pipeline.git.tag >> }
+        - not:
+            equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+  # Uses the same regex as release-tags-workflow
+  release-tags-job: &release-tags-job
+    filters:
+      tags:
+        only: /^[0-9]+\.[0-9]+\.[0-9]+(?!.*-SNAPSHOT).*$/
+      branches:
+        ignore: /.*/
+
 version: 2.1
 
 orbs:
@@ -259,14 +275,12 @@ workflows:
   # to Maven Central and creates a GitHub release for it. The release train has
   # now completed its journey.
   on-release-tag:
-    when:
-      and:
-        - matches: { pattern: "^[0-9]+\\.[0-9]+\\.[0-9]+(?!.*-SNAPSHOT).*$", value: << pipeline.git.tag >> }
-        - not:
-            equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+    <<: *release-tags-workflow
     jobs:
       - publish:
           snapshot: false
           context: maven-central-publishing
+          <<: *release-tags-job
       - create-github-release:
           context: github-bot-public
+          <<: *release-tags-job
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -259,14 +275,12 @@ workflows:
   # to Maven Central and creates a GitHub release for it. The release train has
   # now completed its journey.
   on-release-tag:
-    when:
-      and:
-        - matches: { pattern: "^[0-9]+\\.[0-9]+\\.[0-9]+(?!.*-SNAPSHOT).*$", value: << pipeline.git.tag >> }
-        - not:
-            equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+    <<: *release-tags-workflow
```
tonidero, line 278: I'm wondering if this is needed since we are specifying the `release-tags-job` on each job?

.circleci/config.yml
```diff
@@ -1,3 +1,19 @@
+aliases:
+  # Uses the same regex as release-tags-job
+  release-tags-workflow: &release-tags-workflow
+    when:
+      and:
+        - matches: { pattern: "^[0-9]+\\.[0-9]+\\.[0-9]+(?!.*-SNAPSHOT).*$", value: << pipeline.git.tag >> }
+        - not:
+            equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+  # Uses the same regex as release-tags-workflow
+  release-tags-job: &release-tags-job
```
tonidero, line 10: Hmm I wonder if we should just rename to `release-tags` since it's how we mostly name this for the other SDKs?

.circleci/config.yml
```diff
@@ -1,3 +1,19 @@
...
+    filters:
+      tags:
+        only: /^[0-9]+\.[0-9]+\.[0-9]+(?!.*-SNAPSHOT).*$/
```
tonidero, line 13: I don't think we tag snapshots right? I believe snapshots the way we do them in android/iOS is done automatically on every push to the `main` branch.

tonidero: Also, I wonder if we should just copye the existing `release-tags` alias in iOS/Android, where we basically ignore just some specific tags and release everything else... Mostly to be consistent.

tonidero (commented):
Looks good! Just some questions

.circleci/config.yml
JayShortway: Yea I named it like this to differentiate it from `release-tags-workflow`, which uses the same regex, but needs to use `when` instead of `filters`.

.circleci/config.yml
vegaro: this is cool, we don't do it in other SDKs and I think it makes sense

vegaro (approved):

.circleci/config.yml
JayShortway: We don't tag them indeed, but this was intended to be defensive against the potential case when we push a tag (e.g. for reference purposes) that we don't want to release. This matches all strings that start with a semver, with any suffix except "SNAPSHOT". 

I'm okay to make it less strict to make it behave like iOS/Android, but I think it should still be a single regex for consistency with the workflow filter.

tonidero (approved):

PR #121, opened 2024-06-27
Removes Offerings.getCurrentOfferingForPlacement()

JayShortway:
As the title says. This wasn't implemented on iOS, as it's not exposed by PHC. I noticed in the [docs](https://www.revenuecat.com/docs/getting-started/displaying-products#fetching-offerings-by-placement) that there are no snippets for other hybrids either. So I removed this api for now. We can add it back for the full release if needed.


tonidero (approved):
Yeah, I guess we didn't have time to propagate this API to the hybrids. Probably need to do the PHC part first as you said. For now, I'm ok removing it 👍

PR #122, opened 2024-06-27
Publishes API reference only on release.

JayShortway:
I moved some jobs around in the GHA files to make the API reference publish only on releases. We'll need to keep an eye on this for the next version. 

Note that this still uses GHA to publish the API reference, but imo that's okay for now.


.github/workflows/on_release_tag.yml
```diff
@@ -0,0 +1,127 @@
...
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@417ae3ccd767c252f5661f1ace9f835f9654f2b5 # v3.1.0
+
+    - name: Build libraries
+      run: ./gradlew assemble -x :composeApp:assemble
+    - name: Build sample
+      run: ./gradlew :composeApp:assembleDebug
+
+  generate-api-reference:
+    needs: build
```
tonidero, line 85: Hmm since everything before this is shared with the `main` workflow, is there a way to extract it so we avoid the duplication?

tonidero (approved):
Just a comment but looks good!

.github/workflows/on_release_tag.yml
JayShortway: Totally agree, I think we can do something with [composite actions](https://docs.github.com/en/actions/using-workflows/avoiding-duplication). But I'm not sure how much we'll be using GHA, especially the `main` workflow, long term. So I'd propose to wait and see how that develops and work on organizing the workflows at that point.

tonidero: Makes sense to me 👍 Happy to merge this!

PR #130, opened 2024-07-12
Improves the error message when the PurchasesInitializer is missing from the merged AndroidManifest.xml.

JayShortway:
As the title says. While this should not happen, the previous error message was a bit unhelpful.

### DIFF
--- a/core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/di/AndroidProvider.kt
+++ b/core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/di/AndroidProvider.kt
@@ -56,7 +56,25 @@ internal object AndroidProvider : Application.ActivityLifecycleCallbacks {
 
 internal fun AndroidProvider.requireApplication(): Application =
     application ?: error(
-        "Purchases has no reference to the Application. This is a bug in the library."
+        "Purchases has no reference to the Application. Please make sure the merged " +
+                "AndroidManifest.xml contains the InitializationProvider element like below, " +
+                "with at least the PurchasesInitializer child. This should normally happen " +
+                "automatically." +
+                "\n\n" +
+                """
+        <provider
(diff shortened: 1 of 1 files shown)


core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/di/AndroidProvider.kt
```diff
@@ -56,7 +56,25 @@ internal object AndroidProvider : Application.ActivityLifecycleCallbacks {
...
     application ?: error(
-        "Purchases has no reference to the Application. This is a bug in the library."
+        "Purchases has no reference to the Application. Please make sure the merged " +
+                "AndroidManifest.xml contains the InitializationProvider element like below, " +
+                "with at least the PurchasesInitializer child. This should normally happen " +
+                "automatically." +
+                "\n\n" +
+                """
+        <provider
+            android:name="androidx.startup.InitializationProvider"
```
tonidero, line 66: I'm wondering, if they need to add this to their manifest, whether they would also need to add the androix library to their gradle files...

tonidero (approved):
Do we have any clues why this error might be happening? Seems weird that the manifest isn't merging correctly... If so, we should add a comment explaining the situation or as part of the error message? 

In any case, I think this is ok for now...

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/di/AndroidProvider.kt
JayShortway: Hmm that shouldn't be necessary. The library will be transitively provided by purchases-kmp.

JayShortway:
> Do we have any clues why this error might be happening? 

Unfortunately I don't. I couldn't reproduce it in a fresh project. I checked the merged manifest view in Android Studio, as well as the one that actually ended up in the apk. I will try to get some more info from the reporter.

PR #137, opened 2024-08-07
Remove expect/actual from enums that lack extra properties

JayShortway:
As the title says. Quite some files changed, but the commits are self contained. So I recommend reviewing per commit to make sense of it all. 

---
Closes SDK-3526


tonidero (approved):
LGTM! It is indeed more code, but it will be good to not expose PHC when we finish with these :)

JayShortway:
@tonidero It was actually a net-negative before I added the API dump. Let's see if we can keep that up 😄

PR #140, opened 2024-08-09
Refactor expect/actual from more enums to avoid typealiases

tonidero:
### Description
This continues the work in #137 

In this one, we are removing them from the enums:
- PackageType
- VerificationResult
- BillingFeature
- GoogleReplacementMode
- RecurrenceMode

Additionally, there are a couple extra changes that were made as part of this PR:
- Use `ReplacementMode` in the purchase methods.
- Change `PricingPhase` to use extensions to avoid having to rewrite the whole offering response classes in this PR. I expect we will want to revert this change in future PRs, but did this in order to keep it clean for now.


core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -185,15 +187,18 @@ public actual class Purchases private constructor(private val androidPurchases:
         onSuccess: (storeTransaction: StoreTransaction, customerInfo: CustomerInfo) -> Unit,
         isPersonalizedPrice: Boolean?,
         oldProductId: String?,
-        replacementMode: GoogleReplacementMode,
+        replacementMode: ReplacementMode?,
```
tonidero, line 190: Note this is an additional change I made. I feel it has more sense. I also wonder if we should move to the builder pattern for the purchase methods, as we do in `purchases-android`... But this works for now.

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/models/PricingPhase.android.kt
```diff
@@ -1,5 +1,22 @@
...
-import com.revenuecat.purchases.models.PricingPhase as RcPricingPhase
+import com.revenuecat.purchases.models.PricingPhase as AndroidPricingPhase
+
+public actual typealias PricingPhase = AndroidPricingPhase
+
+public actual val PricingPhase.billingPeriod: Period
+    get() = billingPeriod
+
+public actual val PricingPhase.recurrenceMode: RecurrenceMode
+    get() = recurrenceMode.toRecurrenceMode()
```
tonidero, line 13: I had to move `PricingPhase` to extensions in order to do this.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Package.kt
```diff
@@ -67,17 +67,3 @@ public expect enum class PackageType {
      */
     WEEKLY,
 }
-
-public val PackageType.identifier: String?
```
tonidero, line 71: I think this isn't needed by consumers. I know we expose it from Android (not iOS), but feels like something we should hide there instead.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/BillingFeature.kt
```diff
@@ -1,22 +1,11 @@
...
+public enum class BillingFeature {
     SUBSCRIPTIONS,
     SUBSCRIPTIONS_UPDATE,
     PRICE_CHANGE_CONFIRMATION,
 }
-
-/**
- * Name according to Google Play's `@BillingClient.FeatureType`.
- */
-public val BillingFeature.playBillingClientName: String
```
tonidero, line 16: Also removed this. This shouldn't be needed by consumers IMO. Also noticed we're missing a bunch of enum values here in `purchases-android`... But we can work on that separately.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/GoogleReplacementMode.kt
```diff
@@ -62,13 +62,3 @@ public expect enum class GoogleReplacementMode: ReplacementMode {
      */
     DEFERRED;
 }
-
-public val GoogleReplacementMode.playBillingClientMode: Int
```
tonidero, line 66: Same, not really needed by consumers since we're not using PHC to map these values and we're using `purchases-android`'s purchase methods directly.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/RecurrenceMode.kt
```diff
@@ -21,12 +21,3 @@ public expect enum class RecurrenceMode {
     NON_RECURRING,
     UNKNOWN,
 }
-
-public val RecurrenceMode.identifier: Int?
```
tonidero, line 25: Same here, not needed at the moment.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/SubscriptionOption.kt
```diff
@@ -66,7 +66,7 @@ public val SubscriptionOption.billingPeriod: Period?
  * Not applicable for Amazon subscriptions.
  */
 public val SubscriptionOption.isPrepaid: Boolean
-    get() = fullPricePhase?.recurrenceMode == RecurrenceMode.NON_RECURRING
+    get() = fullPricePhase?.recurrenceMode?.name == RecurrenceMode.NON_RECURRING.name
```
tonidero, line 69: This is actually something I wasn't fully understanding... I was getting a compile time error when comparing the enum values themselves because it was different types of enums... However, the IDE didn't mark this as an error and as I navigated, everything seem to be the same... This should work for now until we actually move `SubscriptionOption` to not use `expect`/`actual` which should avoid this, but if you know what might be happening, it would be great 🙏

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/models/PricingPhase.ios.kt
```diff
@@ -1,9 +1,18 @@
...
-    public actual val billingPeriod: Period,
-    public actual val recurrenceMode: RecurrenceMode,
-    public actual val billingCycleCount: Int?,
-    public actual val price: Price,
-    public actual val offerPaymentMode: OfferPaymentMode?,
-)
+public actual class PricingPhase
+
+public actual val PricingPhase.billingPeriod: Period
+    get() = error("Not available in iOS")
```
tonidero, line 6: Is this ok? Devs should never get an instance of this object in iOS right now... So I thought this was the simplest.

apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/EntitlementInfoAPI.kt
```diff
@@ -47,7 +47,8 @@ private class EntitlementInfoAPI {
             val billingIssueDetectedAtInstant: Instant? = billingIssueDetectedAtInstant
             // FIXME re-enable in SDK-3530
             //  val ownershipType: OwnershipType = ownershipType
-            val verification: VerificationResult = verification
+            // FIXME re-enable in SDK-3530
```
tonidero, line 50: I saw we have had to do this before... I can't say I fully understand it, since the IDE seems to find the correct classes, but it's using different classes when trying to compile the api tests... In any case, I do believe this should be fixed once we move more classes to remove typedefs

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
```diff
@@ -252,7 +251,7 @@ public expect class Purchases {
         onSuccess: (storeTransaction: StoreTransaction, customerInfo: CustomerInfo) -> Unit,
         isPersonalizedPrice: Boolean? = null,
         oldProductId: String? = null,
-        replacementMode: GoogleReplacementMode = GoogleReplacementMode.WITHOUT_PRORATION,
+        replacementMode: ReplacementMode? = null,
```
tonidero, line 254: Note that I also removed the default value from here. That way, we leave it to the native side to choose the default.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/SubscriptionOption.kt
JayShortway: Yea I know what's happening haha:
* The `recurrenceMode` extension is shadowed by the existing member field of the same name.
* The `recurrenceMode` extension is a different type than the existing member field.
* When compiling this for Android `PricingPhase.recurrenceMode` refers to the existing member field on Android, and thus its type is `com.revenuecat.purchases.models.RecurrenceMode`.
* The IDE sees the extension, and the shadowing warning is suppressed (`EXTENSION_SHADOWED_BY_MEMBER`), so the IDE shows `recurrenceMode` as if its type is `com.revenuecat.purchases.kmp.models.RecurrenceMode` (notice the `kmp` package).

These were some of the earliest reported issues in KobanKat haha: 
* #48 
* #51 

I created [SDK-3464](https://linear.app/revenuecat/issue/SDK-3464/detekt-rule-to-error-when-shadows-change-types) to _detekt_ these issues, but if we move away from `expect`/`actual` I _expect_ that this is not needed anymore as we avoid these types of issues.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/models/PricingPhase.ios.kt
JayShortway: Yeah I think that's fine. This is also temporary right?

apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/EntitlementInfoAPI.kt
JayShortway: Yea I think this is the same issue as in the previous comment.

JayShortway (commented):
Nice job converting these! We're well on our way 😄

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/SubscriptionOption.kt
tonidero: Ohhh makes sense thanks for explaining! And yeah, I think when we move away from `expect`/`actual` this sholdn't be a problem, so maybe it's ok to leave as is for now then...

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/models/PricingPhase.ios.kt
tonidero: Yup, once we move away from `expect`/`actual` this shouldn't be an issue.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/SubscriptionOption.kt
JayShortway: Yes I think it's fine to leave it as-is, as long as we don't release it in this state haha

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/models/PricingPhase.android.kt
JayShortway: Nevermind, I'll just revert it for now. Less decision-making for everyone.

JayShortway: PR with the revert opened: https://github.com/RevenueCat/purchases-kmp/pull/141

JayShortway:
@tonidero I set the base branch of this PR to the integration branch, and merged the integration branch in. I know you prefer rebase, but I didn't want to end up having to force push to someone else's branch. Hope you don't mind! 🙏

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -226,15 +233,18 @@ public actual class Purchases private constructor(private val androidPurchases:
...
     ): Unit = androidPurchases.purchaseWith(
         purchaseParams = PurchaseParams.Builder(
             AndroidProvider.requireActivity(),
             subscriptionOption
         ).apply {
             if (isPersonalizedPrice != null) isPersonalizedPrice(isPersonalizedPrice)
             if (oldProductId != null) oldProductId(oldProductId)
-        }.googleReplacementMode(replacementMode)
+            if (replacementMode != null && replacementMode is GoogleReplacementMode) {
```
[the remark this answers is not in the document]
tonidero: Removed in https://github.com/RevenueCat/purchases-kmp/pull/140/commits/9eeba5dab4452bed74f411038def147d418902ed

tonidero:
Thanks for handling that @JayShortway! I believe I addressed most comments. Do you think there is anything else that we should do in this PR?

JayShortway:
@tonidero @vegaro Integration branch is rebased onto `main`, and this PR is up to date with that. After we squash and merge this, history should be clean again. 😄  (I made a note of the number of files and lines changed before all the force-pushes, and it's the same now.)

@tonidero I think this PR is good to go!

JayShortway (approved):

tonidero:
Hmm seems tests are not passing with the recent merge commits 🤔

tonidero:
Ohh ok, tests seem to only run on PRs pointing to `main`... I think it might be ok to merge this then, since tests were passing before.

PR #143, opened 2024-08-09
Reenables #137

JayShortway:
These are the exact same changes as in #137, but targeting the integration branch instead of `main`. 

See for context:
* #137 
* #141


tonidero (approved):
Gave it a quick glance over. Looks good!

PR #145, opened 2024-08-13
[DO NOT MERGE] More StoreTransaction iOS constructors

JayShortway:


core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -264,8 +268,17 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
...
-                else onSuccess()
+                // FIXME: Handle this case better. Maybe use onError()?
+                if (storeTransactionMap == null) error("Expected a non-null RCStoreTransactionMap")
+
+                // TODO: Double check the keys.
+                else onSuccess(
+                    StoreTransaction(
+                        storeTransactionMap["transactionIdentifier"] as String,
+                        storeTransactionMap["productIdentifier"] as String,
+                        storeTransactionMap["purchaseDate"] as NSDate,
```
JayShortway, line 279: Not entirely sure these keys are correct.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -264,8 +268,17 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
             completion = { storeTransactionMap, error ->
                 if (error != null) onError(error.error().toPurchasesErrorOrThrow())
 
-                // TODO: Parse storeTransactionMap to a StoreTransaction object
-                else onSuccess()
+                // FIXME: Handle this case better. Maybe use onError()?
+                if (storeTransactionMap == null) error("Expected a non-null RCStoreTransactionMap")
```
JayShortway, line 272: Is a null map a valid state? I'd guess not. 

If it is, we might want to call `onError()` with an appropriate error message.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
fire-at-will: I believe these are correct except for `purchaseDate` - this is an ISO8601 string that can't be automatically converted to a NSDate. I believe we can use `purchaseDateMillis` which contains a Double with the milliseconds since the 1970 Unix epoch.

The dictionary is defined in the PHC [here](https://github.com/RevenueCat/purchases-hybrid-common/blob/948709df1ba6bbda72b0a5391c008c1ab858dc8f/ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreTransaction%2BHybridAdditions.swift#L15)

fire-at-will (approved):

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
JayShortway: Ah good catch, thanks!

PR #150, opened 2024-08-19
Add InstallmentInfos

fire-at-will:
This PR adds `InstallmentInfos` to the KMP SDK.

### DIFF
--- a/core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/models/SubscriptionOption.ios.kt
+++ b/core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/models/SubscriptionOption.ios.kt
@@ -9,4 +9,5 @@ public actual interface SubscriptionOption {
     public actual val presentedOfferingIdentifier: String?
     public actual val presentedOfferingContext: PresentedOfferingContext?
     public actual val purchasingData: PurchasingData
+    public actual val installmentsInfo: InstallmentsInfo?
 }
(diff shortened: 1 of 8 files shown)


core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/models/SubscriptionOption.ios.kt
```diff
@@ -9,4 +9,5 @@ public actual interface SubscriptionOption {
     public actual val presentedOfferingIdentifier: String?
     public actual val presentedOfferingContext: PresentedOfferingContext?
     public actual val purchasingData: PurchasingData
+    public actual val installmentsInfo: InstallmentsInfo?
```
fire-at-will, line 12: I've added this as an optional for iOS when I know it'll always be null. Do I need to add anything to make sure it's being correctly parsed as `null` when it's being parsed out from PHC?

tonidero: The `SubscriptionOption` object itself should always be null in iOS, so I don't think we need to do any extra parsing.

tonidero (approved):
I believe this should work! Maybe good to let @JayShortway also take a look just in case

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/models/SubscriptionOption.ios.kt
fire-at-will: I just tested making a purchase through a paywall in iOS through the KMP purchase tester app and everything worked correctly, so I think we're probably good here.

JayShortway (approved):
LGTM!

PR #157, opened 2024-08-22
Adds :mappings and :models modules

JayShortway:
As the title says. Since we don't want to generate Dokka docs for internal libraries, I made some changes to minimize duplication. 

- Moved all logic from `LibraryConventionPlugin` to separate functions.
- Renamed `LibraryConventionPlugin` to `PublicLibraryConventionPlugin`.
- Added a new `InternalLibraryConventionPlugin`. 
- Created the new `:mappings` module, and applied the new `InternalLibraryConventionPlugin`.
- Set up the dependencies (`:mappings` depends on PHC. `:core` and `:revenuecatui` depend on `:mappings`.)
- Corrected the CocoaPods config where applicable, setting the correct iOS deployment target and version.  
- Also added the new `:models` module, and applied the `PublicLibraryConventionPlugin`.
- Generated Mermaid graphs to visualize dependencies.


core/core.podspec
```diff
@@ -1,14 +1,14 @@
...
+    spec.version                  = '1.0.0-SNAPSHOT'
     spec.homepage                 = ''
     spec.source                   = { :http=> ''}
     spec.authors                  = ''
     spec.license                  = ''
     spec.summary                  = ''
     spec.vendored_frameworks      = 'build/cocoapods/framework/Purchases.framework'
     spec.libraries                = 'c++'
-    spec.ios.deployment_target = '11.0'
+    spec.ios.deployment_target = '13.0'
```
tonidero, line 11: Ohh this should have been part of the PHC 13 bump I imagine... Do we need to update `main` with this change?

core/build.gradle.kts
```diff
@@ -1,13 +1,16 @@
...
 plugins {
-    id("revenuecat-library")
+    id("revenuecat-public-library")
     alias(libs.plugins.kotlin.cocoapods)
     alias(libs.plugins.codingfeline.buildkonfig)
 }
 
 kotlin {
     sourceSets {
+        commonMain.dependencies {
```
tonidero, line 11: Hmm I'm just wondering, since we will be exposing mapping functions for the android/ios source sets, should we depend on the mappings module for the android/ios source sets only? That way, we keep the `common` source set totally independent of the PHC classes?

tonidero (approved):
Just a comment but looks good!

core/core.podspec
JayShortway: This is on `main` already afaik. I just set the correct value in the `cocoapods { }` block in the `build.gradle.kts`. This `.podspec` file is auto-generated from that `cocoapods { }` block.

core/build.gradle.kts
JayShortway: Great idea! Done here: https://github.com/RevenueCat/purchases-kmp/pull/157/commits/fee60185d03dea5858ccc6efe9ac4c6c976ae3b0 
  
(☝️ also regenerated `revenuecatui.podspec`. Apparently I didn't sync before my last commit.)

vegaro (approved):

JayShortway:
Drafted because we need to add `:models`.

JayShortway:
@tonidero @vegaro I just added 2 commits: 
- The first one adds the `:models` module.
- The second one adds mermaid graphs to each module, generated with a Gradle plugin. (For now we need to manually run `./gradlew createModuleGraph` to regenerate.) It's similar to what Now in Android does, e.g. [here](https://github.com/android/nowinandroid/blob/main/core/domain/README.md). They should be rendered automatically on GitHub. IntelliJ also claims support, but haven't seen that working yet. Let me know if you think this is useful.

mappings/mappings.podspec
```diff
@@ -0,0 +1,50 @@
+Pod::Spec.new do |spec|
```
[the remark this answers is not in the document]
JayShortway: I don't _think_ they are needed for KMP. It's just that they are automatically generated when we apply the `cocoapods` Gradle plugin, and we need to apply that plugin to be able to depend on the PHC pod. There are other ways to depend on PHC, but this is by far the easiest. Maybe we can clean this up in the future.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/StoreProduct.kt
```diff
@@ -0,0 +1,6 @@
+package com.revenuecat.purchases.kmp.models
+
+public expect class StoreProduct
```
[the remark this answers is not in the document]
JayShortway: Ah right, sorry! Commented out the class, so the file and directory structure are there.

PR #160, opened 2024-08-26
Refactor remaining classes to new architecture and move them to `models` and `mappings` modules

tonidero:
### Description

This became a huge PR since it performs the rest of the migration to the new architecture for our models. This mostly:
-Moves remaining classes off `expect/actual` classes (except `Purchases` and `Paywall`) in favor of moving to `common` source set or to interfaces to be implemented in each platform.
- Moves all models to the `models` module and platform specific mapping logic to the `mappings` module.


mappings/src/androidMain/kotlin/com/revenuecat/purchases/kmp/mappings/Offering.android.kt
```diff
@@ -0,0 +1,36 @@
+package com.revenuecat.purchases.kmp.mappings
+
+import com.revenuecat.purchases.kmp.Offering
+import com.revenuecat.purchases.kmp.Package
+import com.revenuecat.purchases.Offering as NativeAndroidOffering
+
+public fun NativeAndroidOffering.toOffering(): Offering =
+    AndroidOffering(this)
+
+public fun Offering.toAndroidOffering(): NativeAndroidOffering =
```
tonidero, line 10: We need to make `Offering` an interface and hold the original object since we need to pass it in for paywalls, and it has the paywall data inside.

mappings/src/androidMain/kotlin/com/revenuecat/purchases/kmp/mappings/SubscriptionOptions.android.kt
```diff
@@ -0,0 +1,13 @@
+package com.revenuecat.purchases.kmp.mappings
+
+import com.revenuecat.purchases.kmp.models.SubscriptionOptions
+import com.revenuecat.purchases.models.SubscriptionOptions as AndroidSubscriptionOptions
+
+internal fun AndroidSubscriptionOptions.toSubscriptionOptions(): SubscriptionOptions =
+    SubscriptionOptions(
+        basePlan = basePlan?.let { AndroidSubscriptionOption(it) },
```
tonidero, line 8: Note that this reverts the work that was done before to avoid calculating each of these at construction time by passing lambdas to the constructor. That didn't work when moving the `SubscriptionOptions` to the `models` module.

So I'm passing the calculated values here. Which is how it works in other hybrids. Lmk if you have concerns with this approach and we can think of a different solution...

core/build.gradle.kts
```diff
@@ -12,7 +12,7 @@ kotlin {
             api(projects.models)
         }
         androidMain.dependencies {
-            api(libs.revenuecat.common)
+            implementation(libs.revenuecat.common)
```
tonidero, line 15: 🙌

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/StoreTransaction.kt
```diff
@@ -0,0 +1,28 @@
...
+     * Only the first productId will be processed by the SDK.
+     */
+    val productIds: List<String>,
+
+    /**
+     * Time the product was purchased, in milliseconds since the epoch.
+     */
+    val purchaseTime: Long,
+) {
+    public companion object
```
tonidero, line 27: I added this to be able to add a static method in the `mappings` module... Maybe there is a better way to organize this but I tried to minimize the changes.

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -28,12 +29,12 @@ internal fun UIKitPaywall(
         modifier = modifier,
         factory = {
             val viewController = if (footer) RCPaywallFooterViewController(
-                offering = options.offering as RCOffering?,
+                offering = options.offering?.toIosOffering() as? RCOffering,
```
tonidero, line 32: This I'm still trying to wrap my head around... The `toIosOffering` returns an object of type `cocoapods.PurchasesHybridCommon.RCOffering`, however, this method expects a `objcnames.classes.RCOffering`. However, doing the cast works fine. Need to figure out the difference between both.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -418,3 +450,28 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
     public actual fun setCreative(creative: String?): Unit =
         iosPurchases.setCreative(creative)
 }
+
+internal fun PurchasesConfiguration.storeKitVersionToUse(): StoreKitVersion {
```
JayShortway, line 454: I'm a little bit torn that this is only `internal` to be able to test it haha. What if we make it `private` to `PurchasesConfiguration` and just set `PurchasesConfiguration.storeKitVersion` to whatever this function calculates, in the constructor? That way we can test it by reading `PurchasesConfiguration.storeKitVersion`. 

E.g. in `PurchasesConfiguration.kt`, something like:

```diff
 public class PurchasesConfiguration private constructor(
     public val apiKey: String,
     public val appUserId: String?,
     public val purchasesAreCompletedBy: PurchasesAreCompletedBy,
     public val userDefaultsSuiteName: String?,
-    public val storeKitVersion: StoreKitVersion,
+    storeKitVersion: StoreKitVersion,
     public val showInAppMessagesAutomatically: Boolean,
     public val store: Store?,
     public val diagnosticsEnabled: Boolean,
     public val dangerousSettings: DangerousSettings,
     public val verificationMode: EntitlementVerificationMode,
     public val pendingTransactionsForPrepaidPlansEnabled: Boolean?
 ) {
+    public val storeKitVersion: StoreKitVersion = storeKitVersionToUse(
+        purchasesAreCompletedBy, 
+        storeKitVersion
+    )
}
```

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -36,7 +59,17 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
...
+                IosPurchases.setLogHandler { level, message ->
+                    val tag = "PurchasesKMP"
+                    when (level) {
+                        RCLogLevelVerbose -> value.v(tag, message ?: "")
+                        RCLogLevelInfo -> value.d(tag, message ?: "")
+                        RCLogLevelDebug -> value.d(tag, message ?: "")
+                        RCLogLevelWarn -> value.w(tag, message ?: "")
+                        RCLogLevelError -> value.e(tag, message ?: "", null)
+                        else -> value.e(tag, "Unrecognized log level", null)
+                    }
```
JayShortway, line 71: Same comment as on the Android side. Maybe this can be non-anonymous in `LogHandler.ios.kt`, and maybe we can keep a reference to the original instance passed, so we can return that in `get`?

JayShortway (commented):
Massive job!! 💪 It was a journey haha, but I'm glad to see this thing reaching a conclusion. Maybe some of the smaller things (removing annotations, correcting docs, some styling) could have been split out in a separate PR, but this one would have been big regardless. Had a few comments and suggestions, but nothing major.

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -200,7 +237,7 @@ public actual class Purchases private constructor(private val androidPurchases:
         }
             .build(),
         onError = { error, userCancelled -> onError(error.toPurchasesError(), userCancelled) },
-        onSuccess = { purchase, customerInfo -> onSuccess(purchase!!, customerInfo.toCustomerInfo()) },
+        onSuccess = { purchase, customerInfo -> onSuccess(purchase!!.toStoreTransaction(), customerInfo.toCustomerInfo()) },
```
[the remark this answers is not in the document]
tonidero: That's a good point... I'm not sure either but seems like this should never be null? cc @vegaro

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/StoreTransaction.kt
tonidero: Yeah, I don't like too much to have these as global functions... But I think it works for now 👍

mappings/src/androidMain/kotlin/com/revenuecat/purchases/kmp/mappings/SubscriptionOptions.android.kt
tonidero: Nice one! Yeah, that makes sense, thanks!

mappings/src/androidMain/kotlin/com/revenuecat/purchases/kmp/mappings/Offering.android.kt
```diff
@@ -0,0 +1,36 @@
...
+    override val sixMonth: Package?
+        get() = wrapped.sixMonth?.toPackage()
+    override val threeMonth: Package?
+        get() = wrapped.threeMonth?.toPackage()
+    override val twoMonth: Package?
+        get() = wrapped.twoMonth?.toPackage()
+    override val monthly: Package?
+        get() = wrapped.monthly?.toPackage()
+    override val weekly: Package?
+        get() = wrapped.weekly?.toPackage()
```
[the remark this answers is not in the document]
tonidero: Good idea! Will do that. I do wonder about changing all of them for consistency... But I think it should be ok since these are unmutable and the others should have easy access.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/PurchasesConfiguration.kt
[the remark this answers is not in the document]
tonidero: Hmm yeah not sure... on one hand `PurchasesConfiguration` is a model, but this one actually has some logic, so it might make sense to keep it in `core` yeah... Will move it back for now.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/PurchasesDelegate.kt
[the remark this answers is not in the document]
tonidero: Right, main reason to move it here is to be able to keep the mapping logic in `mappings`... It's true that it's not really a model, but I think it's better to keep the same style as the other models IMO... Wdyt? Maybe we need to rename the `models` module after all 😅

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/LogHandler.kt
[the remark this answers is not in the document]
tonidero: Same reasoning as with `PurchasesDelegate`. I wanted to keep the same code organization, by keeping the class here and the mapping logic in the `mappings` module.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/SubscriptionOptions.kt
```diff
@@ -3,49 +3,36 @@ package com.revenuecat.purchases.kmp.models
...
      * The default [SubscriptionOption]:
      *   - Filters out offers with "rc-ignore-offer" tag
      *   - Uses [SubscriptionOption] WITH longest free trial or cheapest first phase
      *   - Falls back to use base plan
      */
-    public val defaultOffer: SubscriptionOption?
-        get() = getDefaultOffer()
-
+    public val defaultOffer: SubscriptionOption?,
+    private val subscriptionOptions: List<SubscriptionOption>,
```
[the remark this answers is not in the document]
tonidero: This was removed so does not apply anymore

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/StoreTransaction.kt
```diff
@@ -0,0 +1,26 @@
+package com.revenuecat.purchases.kmp.models
+
+/**
+ * Represents an in-app billing purchase.
+ */
+public data class StoreTransaction(
```
vegaro, line 6: We have a mix of data classes and regular classes. I think we should probably stick to regular classes since data classes adds some functions to the API, and making it a regular class in the future will be breaking

mappings/src/iosMain/kotlin/com/revenuecat/purchases/kmp/mappings/StoreProduct.ios.kt
```diff
@@ -0,0 +1,36 @@
...
+    override val type: ProductType = wrapped.productType().toProductType()
+    override val category: ProductCategory = wrapped.productCategory().toProductCategory()
+    override val price: Price = wrapped.toPrice()
+    override val title: String = wrapped.localizedTitle()
+    override val localizedDescription: String = wrapped.localizedDescription()
+    override val period: Period? = wrapped.subscriptionPeriod()?.toPeriod()
+    override val subscriptionOptions: SubscriptionOptions? = null
+    override val defaultOption: SubscriptionOption? = null
+    override val discounts: List<StoreProductDiscount> = emptyList()
+    override val introductoryDiscount: StoreProductDiscount? = null
```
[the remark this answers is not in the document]
tonidero: Great catch! Thanks!

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Offerings.kt
```diff
@@ -0,0 +1,30 @@
+package com.revenuecat.purchases.kmp
+
+/**
+ * This class contains all the offerings configured in RevenueCat dashboard.
+ * For more info see https://docs.revenuecat.com/docs/entitlements
+ */
+public data class Offerings(
```
vegaro, line 7: same comment here about it being a data class

vegaro: Also, I find it a bit confusing Offerings is a data class and Offering an interface. I think it makes up for a very confusing API tbh, should we aim for consistency and make everything an interface? Or at least related classes. Happy to talk about it

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Package.kt
```diff
@@ -9,14 +9,13 @@ import com.revenuecat.purchases.kmp.models.StoreProduct
...
- * @property offeringIdentifier offering this package was returned from.
  */
-public expect class Package
-
-public expect val Package.identifier: String
-public expect val Package.packageType: PackageType
-public expect val Package.storeProduct: StoreProduct
-public expect val Package.presentedOfferingContext: PresentedOfferingContext
+public interface Package {
+    public val identifier: String
```
vegaro, line 14: I would move the documentation here

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -36,7 +59,17 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
                 ?: IosPurchases.logHandler().toLogHandler().also { _logHandler = it }
             set(value) {
                 _logHandler = value
-                IosPurchases.setLogHandler(value.toRcLogHandler())
+                IosPurchases.setLogHandler { level, message ->
+                    val tag = "PurchasesKMP"
+                    when (level) {
+                        RCLogLevelVerbose -> value.v(tag, message ?: "")
+                        RCLogLevelInfo -> value.d(tag, message ?: "")
```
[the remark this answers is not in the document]
tonidero: Nice catch!

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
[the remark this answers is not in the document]
vegaro: I got you. I think we used to have a null storeTransaction because of deferred upgrades, but we changed it and maybe forgot to update the lambda parameter in `purchaseWith`

```
fun Purchases.purchaseWith(
    purchaseParams: PurchaseParams,
    onError: (error: PurchasesError, userCancelled: Boolean) -> Unit = ON_PURCHASE_ERROR_STUB,
    onSuccess: (purchase: StoreTransaction?, customerInfo: CustomerInfo) -> Unit,
) {
    purchase(purchaseParams, purchaseCompletedCallback(onSuccess, onError))
}
```

```
internal fun purchaseCompletedCallback(
    onSuccess: (purchase: StoreTransaction, customerInfo: CustomerInfo) -> Unit,
    onError: (error: PurchasesError, userCancelled: Boolean) -> Unit,
)
```

I can't find the PR where that changed tho, but yes it feels like it should be non-null

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Offerings.kt
tonidero: I'm ok removing data classes and using only normal classes (It's actually recommended for APIs to avoid breaking changes).

As for interface vs class... My only concern there is increasing the complexity of the project, having to define interfaces for everything and then separate classes implementing them... I'm not opposed however, might be cleaner yeah

mappings/src/androidMain/kotlin/com/revenuecat/purchases/kmp/mappings/LogHandler.android.kt
```diff
@@ -0,0 +1,55 @@
...
+import android.util.Log
+import com.revenuecat.purchases.kmp.LogHandler
+import com.revenuecat.purchases.LogHandler as NativeAndroidLogHandler
+
+public fun NativeAndroidLogHandler.toLogHandler(): LogHandler =
+    (this as LogHandlerWrapper).wrapped
+
+public fun LogHandler.toAndroidLogHandler(): NativeAndroidLogHandler = LogHandlerWrapper(this)
+
+public class DefaultLogHandler: LogHandler {
```
tonidero, line 12: Note that I've had to copy the default log handler. This makes sense, since logs are treated differently in Android/Ios and if we want to be able to have logs in the KMP side. Do lmk if you have other thoughts!

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Offerings.kt
tonidero: From talking in Slack, I've removed most data classes, but I've left them as classes for now. Happy to talk more though!

mappings/src/androidMain/kotlin/com/revenuecat/purchases/kmp/mappings/LogHandler.android.kt
JayShortway: I don't think this will work, because `DefaultLogHandler` is not a `LogHandlerWrapper`, and so `toLogHandler()` will crash when `Purchases.logHandler` is read. 

I have a proposal here, which avoids the need to copy the `DefaultLogHandler` by wrapping the default from the native Android side if `Purchases.logHandler` is read before it is set: https://github.com/RevenueCat/purchases-kmp/pull/162

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Offerings.kt
JayShortway: I converted all `data` classes to regular classes here: https://github.com/RevenueCat/purchases-kmp/pull/163. 

Moving to `interface`s can be done in a separate PR imo? Let's discuss on Slack whether we should do it. Currently I'm thinking maybe we should.

JayShortway (commented):
Nice nice! I think we're nearly there!

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -54,7 +66,32 @@ public actual class Purchases private constructor(private val androidPurchases:
             }
 
         @JvmStatic
-        public actual var logHandler: LogHandler by AndroidPurchases.Companion::logHandler
+        public actual var logHandler: LogHandler
+            get() = AndroidPurchases.Companion.logHandler.toLogHandler()
+            set(value) {
+                AndroidPurchases.Companion.logHandler = object : com.revenuecat.purchases.LogHandler {
```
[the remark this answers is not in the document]
JayShortway: Resolved with #162.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
JayShortway: Resolved with #162.

mappings/src/androidMain/kotlin/com/revenuecat/purchases/kmp/mappings/LogHandler.android.kt
JayShortway: Resolved with #162.

JayShortway (approved):
Looks good from my end! I think the only thing left is whether we're moving to interfaces or not, but that can be done separately.

tonidero:
Will merge this for now. We can address the interface vs class in other PR as you said!

PR #164, opened 2024-08-28
Do not expose phc api

tonidero:
### Checklist

  hybrids

### Description

This Integration branch has the changes in order to be able to hide the PHC API


JayShortway:
@tonidero @vegaro 
I think this can be merged, right? What do you think, squashing or not? 

The intermediate commits are "broken", but it's a giant commit otherwise. Would be a wall you run into every time you check the git history. 😅

tonidero:
I would say normal merge, not squashing. But then, fix the CHANGELOG, so it only shows as one item. Putting it up for review!

JayShortway (approved):
I think this is good to go! Didn't go line by line, as this was all reviewed in separate PRs, but checked the public API in my local checkout and that looks good!

PR #165, opened 2024-08-28
Updates the API dump

JayShortway:
As the title says. Forgot this in my lil contribution PR 😳


tonidero (approved):

tonidero:
> Why does Danger say this increases the size of the repo (at all), if we're removing 52 lines? 🤔

Hmm that's very weird indeed... Maybe it's comparing it to `main`? That would make sense...

JayShortway:
>Maybe it's comparing it to `main`? That would make sense...

Yeah that must be it I think.

PR #170, opened 2024-08-29
Fixes a PaywallListener crash

### LINKED ISSUE #168: Crash after subscription iOS - class RCStoreTransaction cannot be cast to class StoreTransaction
- [x] I have read
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched

**Describe the bug**
Crash of the iOS application just after a subscription on Emulator.
`Uncaught Kotlin exception: kotlin.ClassCastException: class RCStoreTransaction cannot be cast to class com.revenuecat.purchases.kmp.models.StoreTransaction`

1. Environment
    1. Platform:
    2. SDK version: `1.0.0-beta.3` & pod `13.0.1`
    3. OS version: iOS 17.6.1
    4. IDE (e.g. Android Studio, Xcode, Fleet): Xcode
    5. IDE version: 15.4
        - [x] Device
        - [x] Emulator/simulator
    7. Environment:
        - [x] TestFlight
        - [ ] Production
    8. How widespread is the issue. Percentage of devices affected.
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to
   have context, eg. stackoverflow, etc.)

**Additional context**

JayShortway:
As the title says. Also added a listener to the tester app, and was able to reproduce it with that. Verified no more crashes on both Android and iOS.

## Cause
We were unsafely casting, which worked fine before when everything was a type alias. That's no longer the case. `StoreTransaction` was already converted away from a type alias in `1.0.0-beta.3`, and we made it "worse" on `main` by converting the rest. 

## Fix
For some weird reason we have double Kotlin mappings for the same underlying ObjC class. For now, I'm casting one to the other (which works because it's the same thing), and then calling our mapper function. 

I also checked some other casts we do, and they _seem_ okay so far. 

Closes #168

### DIFF
--- a/composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/App.kt
+++ b/composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/App.kt
@@ -29,6 +29,7 @@ fun App() {
                 .fillMaxSize(),
             horizontalAlignment = Alignment.CenterHorizontally
         ) {
+            val loggingListener = rememberLoggingPaywallListener()
             var screen by remember { mutableStateOf<Screen>(Screen.Main) }
 
             when (val currentScreen = screen) {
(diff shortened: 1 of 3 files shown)


tonidero (approved):
Ohh I missed this. Thanks for the fix!!

vegaro (approved):

PR #176, opened 2024-09-03
Fixes an incorrect cast to Package.

JayShortway:
As the title says.


tonidero (approved):
🤦‍♂️ Thanks for the fix!

PR #178, opened 2024-09-04
Tester app shows every Offerings and CustomerInfo property

JayShortway:
Greatly expands the tester app. It now recursively reads and shows every property of both `Offerings` and `CustomerInfo`. (Used this to find https://github.com/RevenueCat/purchases-kmp/pull/176 for instance.) 

The diff is on the bigger side, but the complexity is quite low. To illustrate the changes: 
- There's a new `CollapsibleRow` component.
- There are a lot of other simple components that are variations on `CollapsibleRow` to show all model classes, such as `StoreProduct`. 
- You can now set your API keys and appUserId in `local.properties`, which is already `.gitignore`d. 
- The `MainScreen` switches between configuration and content using `AnimatedContent`. 

**Note** that this is not shipped, so not required to merge before v1, but good to have on `main` regardless.


composeApp/build.gradle.kts
```diff
@@ -81,3 +85,43 @@ android {
...
+            value = project.rootProject.getLocalProperty("revenuecat.appUserId").orEmpty()
+        )
+    }
+    targetConfigs {
+        create("android") {
+            buildConfigField(
+                type = STRING,
+                name = "apiKey",
+                value = project.rootProject
+                    .getLocalProperty("revenuecat.apiKey.google")
```
tonidero, line 107: We probably should add the same in purchases android 😅

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/components/PackageRow.kt
```diff
@@ -0,0 +1,45 @@
...
+@Composable
+internal fun PackageRow(
+    pkg: Package,
+    modifier: Modifier = Modifier,
+    label: String = "package"
+) {
+    CollapsibleRow(
+        collapsedContent = { Text(text = "$label: ${pkg.identifier}") },
+        expandedContent = {
+            Column(modifier = Modifier.padding(start = DefaultPaddingHorizontal)) {
```
tonidero, line 37: Could be nice to add buttons for purchasing the objects that can be purchased (package/subscriptionOption/storeProduct) with the appropriate purchase method. And other utility buttons to test other functions as well.... But I think it can happen in future PRs too!

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/components/PackageRow.kt
```diff
@@ -0,0 +1,45 @@
...
+                Text(text = "type: ${pkg.packageType}")
```
tonidero, line 38: Would it make sense to have some vertical spacing for all these? Maybe with a small delimiter... It feels a bit cramped now... But again, not a blocker.

tonidero (approved):
Nice improvement! Just some suggestions, but I think they are not blockers

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/components/PackageRow.kt
JayShortway: Yea I think that's useful, but indeed probably better in a separate PR.

composeApp/build.gradle.kts
JayShortway: True! 😄

Issue #185, opened 2024-09-08
Namespace collision between :core and :models when uniquePackageNames is enabled


alaegin:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read
  the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched
  for [existing Github issues](https://github.com/RevenueCat/purchases-kmp/issues)

**Describe the bug**
With the release of 1.0.0 the `:models` module was added, however, the namespace in it is the same as in `:core` module. With the `android.uniquePackageNames=true` property this leads to an exception during the compilation of the Android app.

https://github.com/RevenueCat/purchases-kmp/blob/7c60a7131a5b7aa0476078ea7a86f7e6dd59de84/models/build.gradle.kts#L25

```
> Task :app-android:processDevDebugMainManifest FAILED
[com.revenuecat.purchases:purchases-kmp-core-android:1.0.0+13.2.0] /Users/alexey/.gradle/caches/8.8/transforms/61696659ac84c82bf00d9657d836fdb4/transformed/core-release/AndroidManifest.xml Error:
	Namespace 'com.revenuecat.purchases.kmp' is used in multiple modules and/or libraries: com.revenuecat.purchases:purchases-kmp-core-android:1.0.0+13.2.0, com.revenuecat.purchases:purchases-kmp-models-android:1.0.0+13.2.0. Please ensure that all modules and libraries have a unique namespace. For more information, See https://developer.android.com/studio/build/configure-app-module#set-namespace
/Users/alex/StudioProjects/own/xxx/app-android/src/main/AndroidManifest.xml Error:
	Validation failed, exiting

See https://developer.android.com/r/studio-ui/build/manifest-merger for more information about the manifest merger.
```

The solution is to change the `namespace` accordingly to have a module name at the end like in other modules:
https://github.com/RevenueCat/purchases-kmp/blob/7c60a7131a5b7aa0476078ea7a86f7e6dd59de84/mappings/build.gradle.kts#L35

Our temporary fix is to disable `android.uniquePackageNames` property in out project.

1. Environment
    1. Platform: Android
    2. SDK version: 1.0.0
    3. OS version: macOS
    4. IDE (e.g. Android Studio, Xcode, Fleet):
[...]

PR #193, opened 2024-09-11
Moves LogHandler and LogLevel to the :core module

JayShortway:
- Moves `LogHandler` and `LogLevel` to the `:core` module, in the same package. This way nothing needs to be deprecated.

This should be a source compatible change, and _might_ be binary compatible? Not sure. Whatever is removed from the `:models` binary interface is added to `:core`.  

This is one of the steps needed to address #185. This is part of a stack of PRs.


tonidero (approved):
Nice that we don't need the deprecations! As for compatibility, I believe it should be binary compatible as well... It's true that if someone used the `models` library directly for any reason, it wouldn't be, but I think it shouldn't be a problem

PR #195, opened 2024-09-11
Enables uniquePackageNames and sets the namespace for the :models module

JayShortway:
As the title says. This is the final step needed to address #185. This is part of a stack of PRs.


gradle.properties
```diff
@@ -9,3 +9,5 @@ kotlin.code.style=official
 #Android
 android.useAndroidX=true
 android.nonTransitiveRClass=true
+# Enabled to avoid regressions of #185.
+android.uniquePackageNames=true
```
tonidero, line 13: Just to make sure, this wouldn't force consumers to also have this setting enabled right? (I believe it wouldn't but just to check)

tonidero (approved):

vegaro (approved):

gradle.properties
JayShortway: Oh great point! I checked by publishing to mavenLocal and integrating in a new project that has
```properties
android.uniquePackageNames=false
```
It builds and runs fine, so I think we're good.

PR #207, opened 2024-09-17
Migrate GitHub Actions to CircleCI

JayShortway:
We were still using GitHub Actions for a lot of our CI needs. This PR migrates pretty much everything to CircleCI. 

Only 2 things still happen on GitHub Actions after this PR: 
- Submitting our dependency graph to GitHub for vulnerability analysis, visible in the Insights tab above.
- Deploying the SDK reference site to GitHub Pages.

I figured this was alright, as these 2 things are not part of our core workflows, and they're working well and not in our way. Let me know if you disagree.

This does not do any optimizations. I expect things to be as slow as they were before. That will be a follow-up PR. 

**Note:** once this is approved I will update the branch rules and required checks so this can be merged.


tonidero (approved):
LGTM! 💯

PR #214, opened 2024-09-19
Fixes the PurchasesDelegate being deallocated on iOS

### LINKED ISSUE #203: PurchasesDelegate.onCustomerInfoUpdated is not being called on subscription state update
After making a purchase using RevenueCat paywall, the customer info is not being updated (onCustomerInfoUpdated is being called)

```
    private fun configure() {
        Purchases.configure(getPurchasesApiKey())
        Purchases.sharedInstance.delegate = object : PurchasesDelegate {
            override fun onCustomerInfoUpdated(customerInfo: CustomerInfo) {
                onReceived(customerInfo)
            }
```

Explicitly updating customer info from cache using this function works as expected.

```
        Purchases.sharedInstance.getCustomerInfo(
            CacheFetchPolicy.CACHE_ONLY,
            onError = {},
            onSuccess = {
                onReceived(it)
            })
```

compose-plugin = "1.6.11"
kotlin = "2.0.0"
purchases-common = "13.2.0"
purchases-kmp = "1.0.0+13.2.0"

JayShortway:
## Bug
The `PurchasesDelegate` sometimes doesn't get called on iOS.

## Cause
On the iOS platform side, the backing field of `Purchases.delegate`, `Purchases.privateDelegate`, is a [`weak var`](https://github.com/RevenueCat/purchases-ios/blob/b695ac1866f2fd56c7484b53b003cdc70adcd1ca/Sources/Purchasing/Purchases/Purchases.swift#L105). On the multiplatform side, we wrap the actual delegate in `PurchasesDelegateWrapper` to conform to `RCPurchasesDelegateProtocol`. Since that is a private implementation detail, the only reference held to our wrapper is the `weak var`. This means that it gets deallocated the first chance it gets. 

## Fix
We keep a strong reference to our `PurchasesDelegateWrapper`. This matches the Android behavior, which keeps a strong reference on the platform side already.

Fixes #203

### DIFF
--- a/core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
+++ b/core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
@@ -107,6 +107,10 @@ public expect class Purchases {
     /**
      * The delegate is responsible for handling promotional product purchases (App Store only) and
      * changes to customer information.
+     *
+     * **Note:** If your delegate is not a singleton, make sure you set this back to null when
+     * you're done to avoid memory leaks. For instance, if your delegate is tied to a screen, set
+     * this to null when the user navigates away from the screen.
      */
     public var delegate: PurchasesDelegate?
 
--- a/core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/PurchasesDelegate.ios.kt
+++ b/core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/PurchasesDelegate.ios.kt
@@ -15,8 +15,9 @@ import platform.darwin.NSObject
 internal fun RCPurchasesDelegateProtocol.toPurchasesDelegate(): PurchasesDelegate =
     (this as PurchasesDelegateWrapper).wrapped
 
-internal fun PurchasesDelegate.toRcPurchasesDelegate(): RCPurchasesDelegateProtocol =
-    PurchasesDelegateWrapper(this)
+internal fun PurchasesDelegate?.toRcPurchasesDelegate(): RCPurchasesDelegateProtocol? =
+    this?.let { PurchasesDelegateWrapper(it) }
+        .also { PurchasesDelegateStrongReference.delegate = it }
 
 private class PurchasesDelegateWrapper(val wrapped: PurchasesDelegate) :
     RCPurchasesDelegateProtocol,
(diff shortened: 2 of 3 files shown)


core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
```diff
@@ -107,6 +107,10 @@ public expect class Purchases {
     /**
      * The delegate is responsible for handling promotional product purchases (App Store only) and
      * changes to customer information.
+     *
+     * **Note:** If your delegate is not a singleton, make sure you set this back to null when
+     * you're done to avoid memory leaks. For instance, if your delegate is tied to a screen, set
+     * this to null when the user navigates away from the screen.
```
JayShortway, line 113: This already applied to Android, so would already have been useful to call out, but now it applies to both platforms.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/PurchasesDelegate.ios.kt
```diff
@@ -15,8 +15,9 @@ import platform.darwin.NSObject
 internal fun RCPurchasesDelegateProtocol.toPurchasesDelegate(): PurchasesDelegate =
     (this as PurchasesDelegateWrapper).wrapped
 
-internal fun PurchasesDelegate.toRcPurchasesDelegate(): RCPurchasesDelegateProtocol =
-    PurchasesDelegateWrapper(this)
+internal fun PurchasesDelegate?.toRcPurchasesDelegate(): RCPurchasesDelegateProtocol? =
+    this?.let { PurchasesDelegateWrapper(it) }
+        .also { PurchasesDelegateStrongReference.delegate = it }
```
JayShortway, line 20: This is changed so that when the `delegate` is set to null, the strong reference is cleared.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
tonidero: Hmm it's true that in iOS delegates are much more commonly held weakly... But yeah, since this is KMP, I think this makes sense 👍

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/PurchasesDelegate.ios.kt
```diff
@@ -45,3 +46,20 @@ private class PurchasesDelegateWrapper(val wrapped: PurchasesDelegate) :
...
+ * implementation detail, the only reference held to our wrapper is the `weak var`. This means that
+ * it gets deallocated the first chance it gets. For this reason, we keep a strong reference to our
+ * `PurchasesDelegateWrapper` here. This matches the Android behavior, which keeps a strong
+ * reference on the platform side already.
+ *
+ * **Note**: we cannot make our `PurchasesDelegateWrapper` an `object` directly, as objects cannot
+ * extend `NSObject`. See also
+ * [KT67930](https://youtrack.jetbrains.com/issue/KT-67930/Getting-Crash-while-building-KMM-project-with-XCode-15.3#focus=Comments-27-9796215.0-0)
+ */
+private object PurchasesDelegateStrongReference {
```
tonidero, line 63: Hmm it would be great if we could at least tie it to the `Purchases` object lifecycle... So if the `Purchases` object is cleared, this is also cleared... Wdyt?

tonidero (commented):
Just one comment about tying the lifecycle of this to the `Purchases` object. Looks good otherwise!

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/PurchasesDelegate.ios.kt
```diff
@@ -45,3 +46,20 @@ private class PurchasesDelegateWrapper(val wrapped: PurchasesDelegate) :
...
+    var delegate: RCPurchasesDelegateProtocol? = null
```
tonidero, line 64: Would it be possible to just hold this inside the `Purchases.ios` class instead? Just to make sure if it gets dereferenced (without calling the `close` function, we also clear it. (honestly this is an edge case, since usually `Purchases` would live forever, but just in case)

tonidero (approved):
Just one more comment, but looks good!

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/PurchasesDelegate.ios.kt
JayShortway: Ah yes that makes way more sense haha! https://github.com/RevenueCat/purchases-kmp/pull/214/commits/1810b44490c0b0941836fe36c24bf122fbe5703e

tonidero (approved):
Looks great now, thanks for doing the changes :)

PR #221, opened 2024-09-26
Updates Fastlane plugin to 5b2e35c6985e02b5911c53ebe4d071e742e03ccc

JayShortway:
As the title says. Dependabot didn't pick this up for some reason.


vegaro (approved):
I am still unsure why dependabot sometimes doesn't pick these up

PR #227, opened 2024-09-27
Adds actionlint to lint GitHub Actions workflows

JayShortway:
As the title says. Needed, as GitHub doesn't surface errors, and I find myself going back and forth to https://rhysd.github.io/actionlint/ to fix. So I figured it's better to catch these automatically before they're committed.


tonidero (approved):
Oh interesting. I guess it will still let you push, but will just give an error right? In any case, this seems useful 👍

JayShortway:
@tonidero Yeah exactly. I didn't add it to the required checks, because I made it to only run when the GitHub Actions workflows actually change. If we want it to always run, we can add it to the required checks.

PR #234, opened 2024-10-03
Attempts to fix ffi by removing the platform specification.

JayShortway:
## Bug
We've been getting [this error on CircleC](https://app.circleci.com/pipelines/github/RevenueCat/purchases-kmp/729/workflows/cb0011c8-96a8-4bff-b975-22d611beb331/jobs/2015)I:

> Your bundle is locked to ffi (1.17.0-arm64-darwin) from rubygems repository
https://rubygems.org/ or installed locally, but that version can no longer be
found in that source. That means the author of ffi (1.17.0-arm64-darwin) has
removed it. You'll need to update your bundle to a version other than ffi
(1.17.0-arm64-darwin) that hasn't been removed in order to install. 

## Cause
However, 1.17.0 is the latest version at the time of writing. Others have similar issues, e.g. see: https://github.com/ffi/ffi/issues/1105. It is mentioned in that thread that the platform specification in `Gemfile.lock` causes the issue. In our case, it also seems likely that the platform specification causes issues, as the error mentions darwin, but the CIrcleCI job runs in a Docker container.

## Fix
The implemented fix is to downgrade to the version we use in purchases-ios: `1.16.3`. ([1.17.0 is the first version to introduce per-platform binaries.](https://github.com/ffi/ffi/releases/tag/v1.17.0.rc1))

## Alternative fix (not in this PR)
This fix is not in this PR, but documented in case the implemented fix doesn't work. In that case, we can try this next.

1. Remove the any `ffi` lines from the original `Gemfile.lock`, except lines declaring a transitive dependency. 
2. Add [`force_ruby_platform`](https://bundler.io/v2.5/man/gemfile.5.html#FORCE_RUBY_PLATFORM)  to the `Gemfile`.
3. Run `bundle install`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,4 @@ gem 'rb-readline'
 gem 'fastlane'
 gem 'cocoapods'
 
-# Workaround for https://github.com/ffi/ffi/issues/1105
-# https://bundler.io/v2.5/man/gemfile.5.html#FORCE_RUBY_PLATFORM
-gem 'ffi', force_ruby_platform: true
-
 eval_gemfile("fastlane/Pluginfile")
--- a/Gemfile
+++ b/Gemfile
@@ -5,4 +5,8 @@ gem 'rb-readline'
 gem 'fastlane'
 gem 'cocoapods'
 
+# Workaround for https://github.com/ffi/ffi/issues/1105
+# https://bundler.io/v2.5/man/gemfile.5.html#FORCE_RUBY_PLATFORM
+gem 'ffi', force_ruby_platform: true
+
 eval_gemfile("fastlane/Pluginfile")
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -5,4 +5,8 @@ gem 'rb-readline'
 gem 'fastlane'
 gem 'cocoapods'
 
+# Workaround for https://github.com/ffi/ffi/issues/1105
+# https://bundler.io/v2.5/man/gemfile.5.html#FORCE_RUBY_PLATFORM
+gem 'ffi', force_ruby_platform: true
```
tonidero, line 10: Hmm seems iOS is simply using an older version... I'm worried about this adding a explicit dependency which we don't actually need... I wonder if we can just downgrade to the same version we have in iOS?

tonidero (approved):
Just a comment... But maybe it's ok to ship as is

Gemfile
JayShortway: Yea that might be simpler. Should be done now!

Issue #242, opened 2024-10-14
Support for Offerings.getCurrentOfferingForPlacement()


doolle89:
I saw that Offerings.getCurrentOfferingForPlacement(placementId: String) was removed in 1.0.0-beta.2, so I wander if there is some alternative way to use this feature? If not, is it going to be added back at some point and when it could be expected? 

Thanks in advance!

JayShortway:
Hi, thanks for this request! 

It was removed, because it wasn't working on iOS. The intention is indeed to bring it back. I don't think there's another way to use this feature right now, but we'll have a look at what the best way would be to revive it.

doolle89:
Thanks for looking into it so quickly!

JayShortway:
@doolle89 No worries, happy to help! 

This has just been released in 1.2.0+13.6.0. Let me know how it works for you!

doolle89:
@JayShortway I tested 1.2.0+13.6.0, Offerings.getCurrentOfferingForPlacement() works as expected :)

Thanks again!

JayShortway:
@doolle89 Thanks for letting me know, happy to hear!

PR #243, opened 2024-10-16
Adds `Offerings.getCurrentOfferingForPlacement()`

JayShortway:
This adds the `Offerings.getCurrentOfferingForPlacement()` functionality back, which was removed in https://github.com/RevenueCat/purchases-kmp/pull/121. 

In order to make this non-breaking, I added a lambda constructor parameter to `Offerings`. I don't entirely love it, but all the alternatives I could think of are breaking in one way or another (e.g. making `Offerings` an interface). Please do let me know all your ideas! I'm also happy to discuss this more. 

- Dependency: https://github.com/RevenueCat/purchases-hybrid-common/pull/938
- Addresses: https://github.com/RevenueCat/purchases-kmp/issues/242

### DIFF
--- a/apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/OfferingsAPI.kt
+++ b/apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/OfferingsAPI.kt
@@ -5,6 +5,24 @@ import com.revenuecat.purchases.kmp.models.Offerings
 
 @Suppress("unused", "UNUSED_VARIABLE")
 private class OfferingsAPI {
+
+    fun checkConstructor(
+        all: Map<String, Offering>,
+        current: Offering?,
+        getCurrentOfferingForPlacement: (placementId: String) -> Offering?
+    ) {
+        val offerings1 = Offerings(
+            all = all,
+            current = current
+        )
+
+        val offerings2 = Offerings(
+            all = all,
+            current = current,
+            getCurrentOfferingForPlacement = getCurrentOfferingForPlacement,
+        )
+    }
+
     fun check(offerings: Offerings) {
         with(offerings) {
             val current: Offering? = current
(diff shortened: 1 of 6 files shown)


tonidero (commented):
Hmm the alternative is following the API in RN, where this method is in `Purchases`: https://github.com/RevenueCat/react-native-purchases/blob/a34e4ab926254c58c3f53c8a8ea6aced133b5684/src/purchases.ts#L407.... Not sure if I prefer that TBH, to avoid passing the block... Wdyt? Also running out of other ideas without making it a breaking change

JayShortway:
@tonidero Hmm yea I don't think I prefer that either. That's an(other) async API which I think is less ergonomic than just making it available on the `Offerings` type. 

Maybe adding the lambda to the constructor is not super terrible? Devs are unlikely to be constructing it, and even if they are (e.g. in tests), they don't have to come up with an implementation of the lambda as it's optional. 

There's always the option to deprecate the constructor and make it an interface in a major, but then we have an interface, which is not great in itself haha.

tonidero:
Hmm so after looking at the other hybrids, I think all of them are actually using the function in `Purchases`... So I'm thinking, for the sake of consistency, I might prefer that.

FWIW, I think usage would be fine, as in the developer can just call the `Purchases.sharedInstance.getCurrentOfferingForPlacement` wherever they need, since the value should be cached in the native layer, it should be quick...

JayShortway:
@tonidero I think we're either going to be consistent with the native platform SDKs or with the hybrid SDKs. 

I agree that it'll be quick because it's cached, but it is still an async API. Async APIs are far less user-friendly than sync APIs imo. What if a dev wants to get all `Offering`s _and_ get the current one for a placement? 
- In the `Purchases.sharedInstance.getCurrentOfferingForPlacement` case, they'd need to call 2 async APIs (`Purchases.sharedInstance.getCurrentOfferingForPlacement` and `Purchases.sharedInstance.getOfferings`). If they're not using coroutines, they are officially in callback hell. 
- In the `Offerings.getCurrentOfferingForPlacement` case, they'd just need the one call to `Purchases.sharedInstance.getOfferings` and have all they need. 

I think having a slightly odd constructor parameter (that devs are unlikely to call) is a price worth paying for this user-friendliness. 

@vegaro Wanna be the tie breaker here? 😄

vegaro:
I am leaning towards the native way of doing this (`Offerings.getCurrentOfferingForPlacement`). Even if we have to store that lambda in the constructor, I think that's implementation details for a nicer API.

vegaro (approved):

tonidero:
Sounds good then 👍

PR #247, opened 2024-10-17
Fixes iOS PaywallFooter height

JayShortway:
This is a regression introduced by Compose Multiplatform 1.7.0. 

## Bug
The `PaywallFooter` is not laid out correctly. Its height is correct, but it is placed too low in its container. This only happens on iOS. Android is fine. 

## Cause
After some experimentation with safe area insets, I found that it is actually caused by `animateContentSize()`. We are using this, because we don't immediately know the height of the footer. So we set it at 0 at first, and then animate it up once we know it. 

## Fix
Removing the `animateContentSize()` modifier completely fixes the layout. 

## Improvement
I dug a bit deeper, and was actually able to avoid the need for the animation by figuring out the footer height before the first frame is drawn, using `intrinsincContentSize`. This means the footer is immediately of the right height, no animations or janky frames. 

## Comparison

| Before | After |
|--------|-------|
| Simulator Screenshot - iPhone 15 Pro - 2024-10-17 at 19 03 41 | Simulator Screenshot - iPhone 15 Pro - 2024-10-17 at 18 58 15 |

### DIFF
--- a/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallFooter.kt
+++ b/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallFooter.kt
@@ -1,18 +1,15 @@
 package com.revenuecat.purchases.kmp.ui.revenuecatui
 
+import androidx.compose.animation.animateContentSize
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.unit.dp
@@ -25,18 +22,17 @@ public actual fun PaywallFooter(
     options: PaywallOptions,
     mainContent: @Composable ((PaddingValues) -> Unit)?,
 ) {
-    var height by remember { mutableStateOf(0.dp) }
     val paywallComposable = @Composable {
         UIKitPaywall(
             options = options,
             footer = true,
             modifier = Modifier
                 .fillMaxWidth()
+                .animateContentSize()
+                .wrapContentHeight()
                 .clip(
                     RoundedCornerShape(topStart = DefaultCornerRadius, topEnd = DefaultCornerRadius)
-                )
-                .height(height),
-            onHeightChange = { newHeight -> height = newHeight.dp }
+                ),
         )
     }
 
--- a/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallFooter.kt
+++ b/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallFooter.kt
@@ -1,6 +1,5 @@
 package com.revenuecat.purchases.kmp.ui.revenuecatui
 
-import androidx.compose.animation.animateContentSize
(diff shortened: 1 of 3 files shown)


revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallFooter.kt
```diff
@@ -36,7 +35,6 @@ public actual fun PaywallFooter(
                 .clip(
                     RoundedCornerShape(topStart = DefaultCornerRadius, topEnd = DefaultCornerRadius)
                 )
-                .animateContentSize()
```
JayShortway, line 39: This is the fix.

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -41,12 +42,26 @@ internal fun UIKitPaywall(
...
 
             viewController
+                .also {
+                    // The first subview has an actual intrinsic content size. We immediately let
+                    // the parent know, so it can set our height accordingly. This should happen
+                    // before the first frame is drawn.
+                    if (footer) (it.view.subviews.firstOrNull() as? UIView)
+                        ?.getIntrinsicContentSize()
+                        ?.also(onHeightChange)
+                }
```
JayShortway, line 52: This is the improvement.

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallOptionsKtx.kt
```diff
@@ -80,14 +75,4 @@ internal class IosPaywallDelegate(
...
-
-
-    override fun paywallViewController(
-        controller: RCPaywallViewController,
-        didChangeSizeTo: CValue<CGSize>
-    ) {
-        var height: Int? = null
-        memScoped { height = didChangeSizeTo.ptr.pointed.height.toInt() }
-        onHeightChange(height!!)
-    }
```
JayShortway, line 92: This is no longer needed because we set the height based on the `intrinsicContentSize` now. `didChangeSizeTo` fires multiple times with wildly different values, only the last one being correct. `intrinsicContentSize` only fires once, immediately with the right value.

tonidero: Hmm I'm a bit concerned... Would the kotlin side react to changes in the native view? I'm thinking specifically when changing from the "loading" UI to the "loaded" UI, the height is likely to change and not sure if the current alternative will actually update it then?

JayShortway: You're totally right, thanks for catching this! 

This made me dig even deeper, and I now have an even better solution haha. We're now reporting the right height (from UIKit) immediately in the measurement phase. This means we can use `wrapContentHeight()`, which wasn't working before, and also allows us to keep using `animateContentSize()`. The animation is now only used to animate from loading to loaded, when the size changes, not when it first appears (because we already know the right size at that point). https://github.com/RevenueCat/purchases-kmp/pull/247/commits/da77881643968b97277bc4b116429425f34f57eb

Here's what it looks like when the offering needs to load (first one), and when it is already loaded and passed in (second one):

https://github.com/user-attachments/assets/deb877a1-aa93-4d66-9032-2f618c83abba

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -41,12 +76,28 @@ internal fun UIKitPaywall(
...
                 .apply { setDelegate(delegate) }
                 .also { viewControllerWrapper.value = it }
         },
     )
 }
 
+private fun UIView.getIntrinsicContentSize(): Int {
+    var size: Int? = null
+    memScoped { size = intrinsicContentSize.ptr.pointed.height.toInt() }
```
tonidero, line 93: TIL 👍

tonidero (approved):
I think this should work! 👍

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
JayShortway: Kotlin Native at work 😄

PR #289, opened 2025-01-03
Support web redemption links

tonidero:
### Description

This PR adds support for Redemption Links to our KMP SDK.


composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/DeepLinkHandler.kt
```diff
@@ -0,0 +1,18 @@
+package com.revenuecat.purchases.kmp.sample
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
+
+private val _deepLinkFlow = MutableSharedFlow<String>(
```
tonidero, line 7: I didn't really like having all this at the top-level... But if I added all this inside an `object`, it wasn't recognized by iOS for some reason 😕. Considering this is just a test app, I went ahead with this for now.

tonidero (commented):
Ups test failing, need to fix those after my interview

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/DeepLinkHandler.kt
JayShortway: Do you mean in the `iosMain` Kotlin source set or in the Swift code?

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -494,12 +508,42 @@ public actual class Purchases private constructor(private val androidPurchases:
...
+                    "Invalid URL: ${webPurchaseRedemption.redemptionUrl}"
+                )
+            ))
+        }
+        val nativeWebPurchaseRedemption = try {
+            // Replace this with parseAsWebPurchaseRedemption overload
+            // accepting strings once it's available.
+            val intent = Intent(Intent.ACTION_VIEW, Uri.parse(webPurchaseRedemption.redemptionUrl))
+            AndroidPurchases.parseAsWebPurchaseRedemption(intent)
+        } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
```
JayShortway, line 529: Your own tip 😛 
```suggestion
        } catch (_: Throwable) {
```

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
```diff
@@ -99,6 +101,13 @@ public expect class Purchases {
             features: List<BillingFeature> = listOf(),
             callback: (Boolean) -> Unit,
         )
+
+        /**
+         * Given a url string, parses the link and returns a [WebPurchaseRedemption], which can
+         * be used to redeem a web purchase using [Purchases.redeemWebPurchase]
+         * @return A parsed version of the link or null if it's not a valid RevenueCat web purchase redemption link.
```
JayShortway, line 108: Just a suggestion to consistently use `url` instead of `link`, since the parameter is named `url`:

```suggestion
         * Given a url string, parses it and returns a [WebPurchaseRedemption], which can
         * be used to redeem a web purchase using [Purchases.redeemWebPurchase]
         * @return A parsed [WebPurchaseRedemption] or null if it's not a valid RevenueCat web purchase redemption url.
```

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -126,6 +130,19 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
 
         private fun DangerousSettings.toIosDangerousSettings(): IosDangerousSettings =
             IosDangerousSettings(autoSyncPurchases)
+
+        /**
+         * Given a url string, parses the link and returns a [WebPurchaseRedemption], which can
+         * be used to redeem a web purchase using [Purchases.redeemWebPurchase]
+         * @return A parsed version of the link or null if it's not a valid RevenueCat web purchase redemption link.
+         */
+        public actual fun parseAsWebPurchaseRedemption(url: String): WebPurchaseRedemption? {
```
JayShortway, line 139: Just FYI, we didn't make `iosMain` / `androidMain` docs required, just `commonMain`. Can't hurt of course! (Same comment on the "link" vs "url" applies.)

JayShortway (approved):
Nice one! No real comments other than Antonio's, just some suggestions.

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/DeepLinkHandler.kt
JayShortway: Hmm, did you try the `shared` property? Docs are [here](https://kotlinlang.org/docs/native-objc-interop.html#kotlin-singletons).

```swift
MyObject.shared
MyObject.shared.x
```

Also, the kotlin-swift-interopedia is really handy for cases like this. 😄👇 
- [Here's Swift code reading a Kotlin `object`](https://github.com/kotlin-hands-on/kotlin-swift-interopedia/blob/main/kotlin-swift-interopedia-samples/iosApp/iosApp/Classes/ObjectsExample.swift).
- [Here's the Kotlin definition of that `object`](https://github.com/kotlin-hands-on/kotlin-swift-interopedia/blob/main/kotlin-swift-interopedia-samples/shared/src/commonMain/kotlin/com/jetbrains/swiftinteropplayground/classesandinterfaces/Objects.kt)

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/RedeemWebPurchaseListener.kt
```diff
@@ -0,0 +1,55 @@
+package com.revenuecat.purchases.kmp.models
+
+/**
+ * Interface to handle the redemption of a RevenueCat Web purchase.
+ */
+public fun interface RedeemWebPurchaseListener {
+    /**
+     * Result of the redemption of a RevenueCat Web purchase.
+     */
+    public sealed class Result {
```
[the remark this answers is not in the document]
tonidero: That's a good question 😅. So this is already implemented as a sealed class in android (and equivalents in iOS/Flutter/RN)... I implemented that before deciding to move fully away from these...  I think we can probably change it here now, but it's too late on the others, and then we would have a bit of mixed APIs... Wdyt about continuing using sealed class for this to be consistent with our other SDKs APIs?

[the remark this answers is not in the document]
ajpallares: Yes, I suspected that could be it, but didn't want to assume. So, in this case, I think we should maintain it here for consistency

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/DeepLinkHandler.kt
tonidero: Oh that's really helpful thank you!! Still, that doesn't seem to be the issue. The symbol it's not finding in swift is the `DeepLinkHandlerKt`: `error: cannot find 'DeepLinkHandlerKt' in scope`. I'm not sure what's wrong though, according to those examples, seems to be pretty similar 🤔

tonidero: Oh wait dah... It basically doesn't use the file name in that case it seems, and you can just reference the object directly. Nvm then, will fix this, thanks!

tonidero: Fixed in https://github.com/RevenueCat/purchases-kmp/pull/289/commits/d4362edf2e67aa192708f3487e771e9b63dcdb9f

tonidero:
Will merge this. Lmk if you have any further questions and happy to address in follow-up PRs!

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/DeepLinkHandler.kt
JayShortway: Nice!

Issue #126, opened 2024-07-08
WatchOS target


cvb941:
It would be great to also support other Apple targets. Personally, I want watchOS, but there is also tvOS and macOS.

JayShortway:
This is a good suggestion, thanks! We'll investigate to see how much effort it would be to support more Apple targets.

JayShortway:
Note: make sure that the OS-check done as part of Win-Back Offers remains correct. https://github.com/RevenueCat/purchases-kmp/pull/291#discussion_r1907805702

PR #291, opened 2025-01-08
Support Redeeming iOS Win-Back Offers on Developer Paywalls

fire-at-will:
### Motivation
This PR makes it possible for developers to fetch win-backs that a subscriber is eligible for for a given product or package, and then purchase them from their custom paywalls.

### PHC Dependency
This PR depends on https://github.com/RevenueCat/purchases-hybrid-common/pull/1008, and should not be merged until that PR has been deployed and the new PHC version adopted in the KMP SDK. This branch will not build until that change has been taken in. Please review that PHC PR when reviewing this PR.

#### New Interfaces
This PR introduces a new `WinBackOffer` interface, which contains a `StoreDiscount`:
```kotlin
public interface WinBackOffer {
    public val discount: StoreProductDiscount
}
```

#### New Public Functions
It also introduces four new functions, each of which are only implemented on iOS:
```kotlin
public fun getEligibleWinBackOffersForProduct(
    storeProduct: StoreProduct,
    onError: (error: PurchasesError) -> Unit,
    onSuccess: (List<WinBackOffer>) -> Unit,
)

public fun getEligibleWinBackOffersForPackage(
    packageToCheck: Package,
    onError: (error: PurchasesError) -> Unit,
    onSuccess: (List<WinBackOffer>) -> Unit,
)
    
public fun purchase(
    storeProduct: StoreProduct,
    winBackOffer: WinBackOffer,
    onError: (error: PurchasesError, userCancelled: Boolean) -> Unit,
    onSuccess: (transaction: StoreTransaction, customerInfo: CustomerInfo) -> Unit,
)

public fun purchase(
    packageToPurchase: Package,
    winBackOffer: WinBackOffer,
    onError: (error: PurchasesError, userCancelled: Boolean) -> Unit,
    onSuccess: (transaction: StoreTransaction, customerInfo: CustomerInfo) -> Unit,
)
```

The idea here is that a developer will first fetch the eligible win-back offers for a product/package with one of the `getEligibleWinBackOffers` functions, and display those offers on their paywall, and then redeem it with one of the new `purchase` functions. This mirrors the API design in the other hybrids and the native iOS SDK.

The PR also introduces suspending variations of these APIs, specifically versions with:
- Plain coroutines
- Coroutines + `Either`
- Coroutines + `Result`

#### New `DiscountType` Cases
Adds a new `WINBACK` case to the `DiscountType` enum

#### Purchase Tester Additions
This PR adds a new screen to test the various win-back functions, accessible from a new button on the main screen.


core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -446,4 +605,12 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
 
     public actual fun setCreative(creative: String?): Unit =
         iosPurchases.setCreative(creative)
+
+    private fun isIOSVersion18OrAbove(): Boolean {
+        return currentOSVersion() > "18.0.0"
```
fire-at-will, line 610: It's necessary for us to check the OS version here. Despite the native functions being marked as available on iOS 18.0+, macOS 15.0+, tvOS 18.0+, watchOS 11.0+, and visionOS 2.0+, KMP allows us to call these functions regardless of the OS versions 🤠

Other hybrids skirt around this issue in two ways:
- They perform an availability check in their Obj-C/Swift wrappers
- PHC also performs an availability check

However, since we're calling the native functions directly from Kotlin, we can't take advantage of those mechanisms. This _technically_ works on iOS, but it's not great for a few reasons:
- It depends on the `UIDevice` class, which isn't available in all OS's
- It's only checking for the iOS version, ignoring the OS requirements for other OS's.

Do you happen to know of a better way to handle this @JayShortway?

iosApp/iosApp/SKConfig.storekit
```diff
@@ -0,0 +1,164 @@
+{
+  "appPolicies" : {
```
fire-at-will, line 2: Adding the SKConfig file allows us to test win-back offers on products :)

apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/PurchasesCommonAPI.kt
```diff
@@ -295,4 +295,58 @@ private class PurchasesCommonAPI {
...
+                    onError = { error -> },
+                    onSuccess = { eligibleWinBackOffers ->
+                        val winBackOffer = eligibleWinBackOffers.first()
+
+                        // Purchase the product with the win-back offer
+                        Purchases.sharedInstance.purchase(
+                            storeProduct = product,
+                            winBackOffer = winBackOffer,
+                            onError = { error -> },
+                            onSuccess = { transaction, customerInfo -> }
```
JayShortway, line 319: In other SDKs we should define the types here instead of relying on type inference, as type inference won't catch it if we accidentally change the type. It is not strictly necessary in KMP because it uses the [binary-compatibility-validator](https://github.com/Kotlin/binary-compatibility-validator) to enforce this. But maybe it's a good idea to do it for consistency? 

I'm talking about doing things like this: 
```diff
-            onError = { error -> },
-            onSuccess = { products ->
-                val product = products.first()
+            onError = { error: PurchasesError -> },
+            onSuccess = { products: List<StoreProduct> ->
+                val product: StoreProduct = products.first()
// etc.
```

This would also apply to `checkRedeemingWinBackOffersForPackage()` below.

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/App.kt
```diff
@@ -67,6 +69,8 @@ fun App() {
                         )
                     }
                 }
+
+                is Screen.WinBackTesting -> WinBackTestingScreen()
```
JayShortway, line 73: Love this!

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/MainScreen.kt
```diff
@@ -53,6 +53,7 @@ import kotlinx.coroutines.flow.onStart
 @Composable
 fun MainScreen(
     onShowPaywallClick: (offering: Offering?, footer: Boolean) -> Unit,
+    setScreen: (Screen) -> Unit,
```
JayShortway, line 56: Naming suggestion: `navigateTo`.

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/MainScreen.kt
```diff
@@ -53,6 +53,7 @@ import kotlinx.coroutines.flow.onStart
 @Composable
 fun MainScreen(
     onShowPaywallClick: (offering: Offering?, footer: Boolean) -> Unit,
```
JayShortway, line 55: Would be nice to remove this one and just use `setScreen`/`navigateTo` for the paywall too. I think that should be possible?

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/WinBackTestingScreen.kt
```diff
@@ -0,0 +1,167 @@
...
+
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .verticalScroll(rememberScrollState())
+            .padding(16.dp)
+    ) {
+        Text("Use this screen to fetch eligible win-back offers, purchase products without a win-back offer, and purchase products with an eligible win-back offer.")
+        
+        Text("This test relies on products and offers defined in the SKConfig file, so be sure to launch the PurchaseTester app from Xcode with the SKConfig file configured.")
```
JayShortway, line 26: This is great! I love a tester app that explains on-screen how it should be used 😗👌

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
JayShortway: As of now, our KMP SDK only supports iOS. (See our configured compilation targets [here](https://github.com/RevenueCat/purchases-kmp/blob/main/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt#L21).) So I think that avoids those downsides? 

(I would like the KMP SDK to support more Apple targets in the future. When we do, we can use `expect`/`actual` to make sure this works on all Apple targets. There's https://github.com/RevenueCat/purchases-kmp/issues/126 to track it.)

The comparison might be a bit brittle, as it's comparing strings, but I think it should work (Unlikely, but if `currentOSVersion` ever prefixes a 0, it won't work anymore.) 

Although shouldn't it be this? 
 ```suggestion
         return currentOSVersion() >= "18.0.0"
```

**Review by JayShortway — APPROVED** [review]
Really nice! Adding this to the tester app is amazing! Just had some minor comments.

File: apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/PurchasesCommonAPI.kt (hunk 3 above)
Comment (fire-at-will) [reply apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/PurchasesCommonAPI.kt]: Love this! Added in 03a3acf63ae059e9d2b1ea0870d96477cc8bbe85

File: composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/MainScreen.kt (hunk 5 above)
Comment (fire-at-will) [reply composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/MainScreen.kt]: Good idea! I did this and some other cleanup to the PurchaseTester's navigation in 053f0dee6a07b3d2a798fb3b578ef9265d063439 and 760929fff50ffa8a50881f7e0fa8a1c06da11181

File: composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/MainScreen.kt (hunk 6 above)
Comment (fire-at-will) [reply composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/MainScreen.kt]: That definitely makes things more concise! I've done this and a little more navigational clean up in 760929fff50ffa8a50881f7e0fa8a1c06da11181

File: core/api/core.api [hunk 8]
```diff
@@ -45,8 +47,10 @@ public final class com/revenuecat/purchases/kmp/Purchases {
 	public final fun logIn (Ljava/lang/String;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
 	public final fun logOut (Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;)V
 	public final fun purchase (Lcom/revenuecat/purchases/kmp/models/Package;Lcom/revenuecat/purchases/kmp/models/PromotionalOffer;Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function2;)V
+	public final fun purchase (Lcom/revenuecat/purchases/kmp/models/Package;Lcom/revenuecat/purchases/kmp/models/WinBackOffer;Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function2;)V
```
fire-at-will, line 50: I noticed that on these purchase functions, it's labeling them as `Function2` since I changed the API here midway through development to include the `userCancelled` bool in the `onError` completion handlers. I don't _think_ this will cause any issues, but just wanted to call it out in case.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
```diff
@@ -411,6 +412,74 @@ public expect class Purchases {
...
+     * met.
+     *
+     * @param packageToCheck: The package to check for eligible win-back offers on.
+     * @param onError Will be called if an error occurs, providing a [PurchasesError] describing
+     * what went wrong.
+     * @param onSuccess Will be called if the function completes successfully, including details
+     * on the [WinBackOffer]s that the subscriber is eligible for on the provided package.
+     */
+    public fun getEligibleWinBackOffersForPackage(
+        packageToCheck: Package,
```
fire-at-will, line 442: In the JS hybrids, this parameter is named `aPackage`, but I went with `packageToCheck` here since it felt more in line with the `packageToPurchase` parameter from the purchase package functions.

fire-at-will:
Hey @JayShortway, I've made a _ton_ of improvements here, enough that it's definitely enough to warrant a re-review 😅

Here's what's new since you looked at it last:

### Core Purchasing Functionality
- Modified the `purchase` APIs to include the `userCancelled` Boolean in the `onError` completion handlers
- We've updated to use PHC 13.16.2, which includes the `RCPurchaseParams.noOP` function, which means that we can now properly build the `RCPurchaseParams` object in `Purchases.ios.kt` 🤩

### New Coroutine APIs
Added suspending variations of all four new functions, including:
- Plain coroutines
- Coroutines + `Either`
- Coroutines + `Result`

### Tests
- Updated the `PurchaseCommonAPI` test class to include the new `userCancelled` Boolean
- Updated the `PurchaseCommonAPI` test class to include the new coroutine functions
- Added a new `WinBackOfferAPI` class to the API testers
- Refactored navigation in the Purchase Tester app to simplify it, in the way that you suggested in your first review

### Docs
- Minor tweaks to the JavaDocs on the new functions

Let me know if you have any questions! 🙌

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ktx/Coroutines.kt
```diff
@@ -358,6 +359,127 @@ public suspend fun Purchases.awaitPurchase(
...
+ * these requirements aren't met.
+ *
+ * @param storeProduct: The product to check for eligible win-back offers on.
+ *
+ * @throws PurchasesException in case of an error.
+ *
+ * @return A [List] of eligible [WinBackOffer]s.
+ */
+@Throws(PurchasesException::class, CancellationException::class)
+public suspend fun Purchases.awaitGetEligibleWinBackOffersForProduct(
```
JayShortway, line 375: I don't think we're very consistent with this, but maybe this can be `awaitEligibleWinBackOffersForProduct()`? That way we avoid having 2 verbs. (This comment also applies to the `*Package()` function below, and the `*Either()` and `*Result() variants.)

JayShortway (approved):
Really great job! Amazingly thorough with all the suspend variations of the functions, and with testing this! 💪

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -330,6 +338,212 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
...
+        }
+
+        iosPurchases.eligibleWinBackOffersForProduct(
+            product = storeProduct.toIosStoreProduct(),
+            completion = { eligibleWinBackOffers, error ->
+
+                if (error != null) {
+                    try {
+                        onError(error.toPurchasesErrorOrThrow())
+                    } catch(e: Exception) {
```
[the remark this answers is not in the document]
fire-at-will: Yeah, that feels like a safer option! Would definitely be a good thing to refactor in another PR 👍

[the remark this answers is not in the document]
fire-at-will: Good idea, done in 63c4cc57892e586b283e5cb2daf5bef7130e4a5c!

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ktx/Coroutines.kt
fire-at-will: I think this makes sense and has precedent! Both `awaitPromotionalOffer` and `awaitOfferings` drop the `get` prefix. Will make the change!

fire-at-will: Done in 28489be4e78059e5026ae86fd145619ac55e0608! 😄

JayShortway (approved):
Thanks for making the changes! 🙏

PR #301, opened 2025-01-28
Sets the deployment target of the UI module to 15.

JayShortway:
## Description
I noticed that the deployment target of `purchases-kmp-ui` was set to the same value (13) as the other modules, even though the underlying iOS module is only available for 15 and up. This fixes that.

This also seems to make building the sample app more reliable, at least on my machine.


tonidero (approved):
LGTM!

Issue #309, opened 2025-02-01
Is there a way to check if user is eligible for a trial on iOS?


MrJohn10:
I want to check if the user is eligible for trial for a given offering. Is there a way to call `checkTrialOrIntroDiscountEligibility ` from the iOS sdk or another way I can get that data?

this is my code right now.:
```
        return when (getPlatform().os) {
            OS.Android -> {
                if (offering?.availablePackages?.any { it.storeProduct.subscriptionOptions?.freeTrial != null } == true) {
                    offering
                } else {
                    null
                }
            }

            OS.Ios -> {
                // todo
            }
        }
```

JayShortway:
Hi, thanks for raising this! It seems like this api is indeed missing from purchases-kmp. We'll work on getting it added, and keep this issue updated. Stay tuned!

JayShortway:
This will be available in the next release, which is planned for either today or early next week.

PR #316, opened 2025-02-10
Adds `checkTrialOrIntroPriceEligibility()`

JayShortway:
## Description
This API was missing from purchases-kmp. 

Closes https://github.com/RevenueCat/purchases-kmp/issues/309


core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -312,6 +313,17 @@ public actual class Purchases private constructor(private val androidPurchases:
         )
     }
 
+    public actual fun checkTrialOrIntroductoryPriceEligibility(
```
tonidero, line 316: Hmm should we call this `checkTrialOrIntroPriceEligibility`? A bit shorter.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/IntroEligibility.kt
```diff
@@ -0,0 +1,25 @@
+package com.revenuecat.purchases.kmp.models
+
+/**
+ * Holds the introductory price status. iOS only.
+ */
+public class IntroEligibility(
```
tonidero, line 6: Hmm I wonder if this could be an enum directly? Not sure if we need to hold the enum inside...

tonidero (approved):
Just a couple comments but looks good!

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/IntroEligibility.kt
JayShortway: Yea I'm okay with that. I went for consistency with the other platforms. Or should we avoid the enum altogether and have some predefined constants? 🤔

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
JayShortway: I like it!

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/IntroEligibility.kt
tonidero: Right enums... I would say let's keep it consistent for now then and leave it as you have it? and we can start adapting new APIs in the future?  I'm ok if we want to start doing it here, but might be weird to have different APIs in the different hybrids...

JayShortway: Yea true. Left the enum in for now. https://github.com/RevenueCat/purchases-kmp/pull/316/commits/f1825bd3746cb682446160f1be215e4205e0c267

apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/PurchasesCommonAPI.kt
```diff
@@ -375,6 +380,18 @@ private class PurchasesCommonAPI {
         val handler = Purchases.logHandler
     }
 
+    fun checkFetchingTrialOrIntroductoryPriceEligibility(
```
tonidero, line 383:
```suggestion
    fun checkFetchingTrialOrIntroPriceEligibility(
```

tonidero (approved):
LGTM!

apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/PurchasesCommonAPI.kt
JayShortway: Good catch! https://github.com/RevenueCat/purchases-kmp/pull/316/commits/9ee19b8629a084863461fb14a0c1a6022a38194e

PR #326, opened 2025-02-21
Fix crash when mapping store type RCBilling to KMP

tonidero:
### Description
While testing web redemptions, I noticed we could crash in iOS if the user had an RCBiling/Web purchase.


JayShortway (approved):
Good one!

PR #340, opened 2025-03-06
Avoids unnecessary scrolling delay on iOS paywalls

### LINKED ISSUE #334: iOS Paywall v2 scrolling latency
There is a slight delay in scrolling on iOS.
I think that's because we use `UIKitInteropInteractionMode.Cooperative` instead of `UIKitInteropInteractionMode.NonCooperative` in the `UIKitViewController`. But that's just a guess.

compose-plugin = "1.7.3"
purchases-kmp = "1.6.0+13.19.0"

JayShortway:
## Description
By default, Compose and UIKit interact in `Cooperative` mode. This means that for every scroll action, Compose is given the chance to react to it, before passing that on to the underlying UIKit (and in our case SwiftUI) view. This adds a noticeable delay to the scroll action, which is unnecessary in our case. We only ever need UIKit to respond to the scroll.

## Fix
The fix is to use the `NonCooperative` mode. Unfortunately this is an experimental API. Since we would like to avoid binary incompatibility issues, this PR adds a best-effort approach. If the experimental API has changed in a newer CMP version that we happened to be linked against at runtime, we use the default (`NonCooperative`) mode. This would avoid a crash, at the cost of reintroducing the scrolling delay.  

Fixes #334.

### DIFF
--- a/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
+++ b/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
@@ -5,10 +5,12 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
+import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.layout.layout
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.unit.dp
+import androidx.compose.ui.viewinterop.UIKitInteropProperties
 import androidx.compose.ui.viewinterop.UIKitViewController
 import cocoapods.PurchasesHybridCommonUI.RCPaywallFooterViewController
 import cocoapods.PurchasesHybridCommonUI.RCPaywallViewController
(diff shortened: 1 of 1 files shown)


tonidero (approved):
Thanks for fixing this!

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -103,3 +106,28 @@ private fun UIView.getIntrinsicContentSizeOfFirstSubView(): Int? =
...
+
+/**
+ * Uses NonCooperative UIKitInteropInteractionMode if available, for snappy scrolling.
+ */
+@OptIn(ExperimentalComposeUiApi::class)
+private fun nonCooperativeUiKitInteropPropertiesNonExperimental(): UIKitInteropProperties =
+    try {
+        // This UIKitInteropProperties constructor and the UIKitInteropInteractionMode type are
+        // experimental at the time of writing. We try to use them, but this could fail if we're
+        // linked against a newer version of CMP that changed this API. If so, we fall back to the
```
ajpallares, line 118: I think you meant `KMP`?
```suggestion
        // linked against a newer version of KMP that changed this API. If so, we fall back to the
```

ajpallares (approved):
Great!

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
JayShortway: Good eye, but I actually mean CMP, for Compose Multiplatform 😄 That's the library that this API is coming from.

JayShortway: Yes indeed, better safe than sorry! Thanks for paying attention 😄

PR #341, opened 2025-03-06
Only runs tests once on release branches, instead of twice.

JayShortway:
## Description
I noticed a while back that 2 workflows kick off on release branches (`on-any-branch` and `on-release-branch`), both running the full test suite. This PR fixes that.


tonidero (approved):
Looks good!

ajpallares (approved):
Makes sense!

PR #353, opened 2025-03-13
Attempts to fix ffi by removing the platform specification.

JayShortway:
## Description
[#352](https://github.com/RevenueCat/purchases-kmp/pull/352) reintroduced an [issue](https://app.circleci.com/pipelines/github/RevenueCat/purchases-kmp/1230/workflows/3c8da019-71ea-4a5e-b385-f47994fbd7b0/jobs/6619) where `ffi` specifies a platform that does not match with CI. We've seen this before in https://github.com/RevenueCat/purchases-kmp/pull/234. Back then we opted to downgrade, but I'm proposing to implement the `force_ruby_platform` workaround from that PR's [first commit](https://github.com/RevenueCat/purchases-kmp/pull/234/commits/8572b8ff50790ededcd002992d4a726f9f636313) now, as we otherwise need to keep remembering to downgrade `ffi`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,4 +5,8 @@ gem 'rb-readline'
 gem 'fastlane'
 gem 'cocoapods'
 
+# Workaround for https://github.com/ffi/ffi/issues/1105
+# https://bundler.io/v2.5/man/gemfile.5.html#FORCE_RUBY_PLATFORM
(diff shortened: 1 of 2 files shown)


tonidero (approved):
Ohhh I totally missed that. I think this makes sense 👍

PR #373, opened 2025-04-11
Enables accessibility features in UIKit views.

### LINKED ISSUE #372: iOS: No voice over support on paywalls v1 and v2
I've noticed the voice over doesn't work at all on the paywalls.
revcat version: 1.7.4+13.28.0

JayShortway:
## Description
Closes #372

### DIFF
--- a/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitInteropPropertiesKtx.kt
+++ b/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitInteropPropertiesKtx.kt
@@ -6,7 +6,10 @@ import androidx.compose.ui.viewinterop.UIKitInteropProperties
 /**
  * Set UIKitInteropProperties without having to opt-in to experimental APIs.
  *
- * @param nonCooperativeInteractionMode Uses NonCooperative UIKitInteropInteractionMode if available, for snappy scrolling.
+ * @param nonCooperativeInteractionMode Uses NonCooperative UIKitInteropInteractionMode if
+ * available, for snappy scrolling.
+ * @param isNativeAccessibilityEnabled Enables the native accessibility features of the underlying
+ * UIKit view.
  */
 @OptIn(ExperimentalComposeUiApi::class)
 internal fun uiKitInteropPropertiesNonExperimental(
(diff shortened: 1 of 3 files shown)


revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitInteropPropertiesKtx.kt
```diff
@@ -4,26 +4,38 @@ import androidx.compose.ui.ExperimentalComposeUiApi
...
+ * UIKit view.
  */
 @OptIn(ExperimentalComposeUiApi::class)
-internal fun nonCooperativeUiKitInteropPropertiesNonExperimental(): UIKitInteropProperties =
-    try {
+internal fun uiKitInteropPropertiesNonExperimental(
+    nonCooperativeInteractionMode: Boolean,
+    isNativeAccessibilityEnabled: Boolean,
+): UIKitInteropProperties =
+    if (nonCooperativeInteractionMode) try {
```
ajpallares, line 19: I'm guessing this is for future proofing in case we don't want `nonCooperativeInteractionMode` in the future, right? Because I see that it's always passed to `true`

ajpallares (approved):
Thank you for fixing this! It looks good to me! Only a question, just out of curiosity

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitInteropPropertiesKtx.kt
```diff
@@ -4,26 +4,38 @@ import androidx.compose.ui.ExperimentalComposeUiApi
...
+ * available, for snappy scrolling.
+ * @param isNativeAccessibilityEnabled Enables the native accessibility features of the underlying
+ * UIKit view.
  */
 @OptIn(ExperimentalComposeUiApi::class)
-internal fun nonCooperativeUiKitInteropPropertiesNonExperimental(): UIKitInteropProperties =
-    try {
+internal fun uiKitInteropPropertiesNonExperimental(
+    nonCooperativeInteractionMode: Boolean,
+    isNativeAccessibilityEnabled: Boolean,
```
tonidero, line 17: Wondering, should we set default values of these to `true`? Since that's probably what we would like most of the time.

tonidero (approved):
Just a comment but FFTI. LGTM!

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitInteropPropertiesKtx.kt
JayShortway: Yea, I intended for this to be more of a general utility function, and didn't want to make any assumptions about the call site. The behavior is quite different and dependent on what kind of (UIKit) view we're working with. Having these parameters makes it a bit more versatile.

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitInteropPropertiesKtx.kt
tonidero: Sounds good, thanks for explaining!

PR #389, opened 2025-05-05
Add `getStorefront` API

tonidero:
### Description
Adds new API to obtain storefront information like the store account country code


composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/components/CustomerInfoSection.kt
```diff
@@ -23,6 +30,14 @@ internal fun CustomerInfoSection(
...
     modifier: Modifier = Modifier
 ) {
+    var storefront by remember {
+        mutableStateOf<Storefront?>(null)
+    }
+    LaunchedEffect(Unit) {
+        Purchases.sharedInstance.getStorefront { result ->
+            storefront = result
+        }
+    }
```
JayShortway, line 40: Nit, we can use [`produceState()`](https://developer.android.com/develop/ui/compose/side-effects#producestate) here.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/Storefront.kt
```diff
@@ -0,0 +1,9 @@
+package com.revenuecat.purchases.kmp.models
+
+/**
+ * Represents the storefront information for the current store account.
```
JayShortway, line 4:
```suggestion
 * Represents the storefront information for the user's current store account.
```

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
```diff
@@ -135,6 +136,8 @@ public expect class Purchases {
      */
     public val store: Store
 
+    public fun getStorefront(callback: (storefront: Storefront?) -> Unit)
```
JayShortway, line 139: Should we add a `suspend` variant of this one as well?

JayShortway (approved):
Nice, thanks for adding this!

PR #392, opened 2025-05-07
Bump Kotlin to 2.1.10 and JetBrains Compose to 1.8.0 stable

skydoves:
### Motivation

Bump Kotlin to 2.1.10 and JetBrains Compose to 1.8.0 stable.


build.gradle.kts
```diff
@@ -101,11 +102,6 @@ allprojects {
 
 apiValidation {
     ignoredProjects.addAll(listOf("apiTester", "composeApp", "mappings"))
-
-    @OptIn(kotlinx.validation.ExperimentalBCVApi::class)
-    klib {
-        enabled = true
-    }
```
JayShortway, line 108: Is klib support automatically enabled in bcv now?

gradle/libs.versions.toml
```diff
@@ -30,8 +30,9 @@ android-application = { id = "com.android.application", version.ref = "agp" }
 android-library = { id = "com.android.library", version.ref = "agp" }
 codingfeline-buildkonfig = { id = "com.codingfeline.buildkonfig", version = "0.15.1" }
 iurysouza-modulegraph = { id = "dev.iurysouza.modulegraph", version = "0.10.0" }
-jetbrains-compose = { id = "org.jetbrains.compose", version = "1.7.0" }
+jetbrains-compose = { id = "org.jetbrains.compose", version = "1.8.0" }
```
JayShortway, line 33: I don't believe anything is broken for library consumers right now, even if they are on Kotlin 2.1.10 and CMP 1.8.0. Is there anything that would be improved by us moving to those versions? I suppose Xcode 16 support is nice for ourselves haha.

JayShortway (commented):
Thanks for doing this, this is great! I'll hold off on approving until we actually want to merge this, as an argument can be made that this should be a major version update. We'll probably do one in June when BC8 gets released.

JayShortway:
I created a Linear in our "Next majors" project and linked this PR: [SDK-3763](https://linear.app/revenuecat/issue/SDK-3763/tentative-update-to-kotlin-21-and-cmp-180-in-purchases-kmp).

build.gradle.kts
skydoves: Ops! I removed this by mistake for testing to bump up Kotlin versions. Recovered in the additional commit.

gradle/libs.versions.toml
skydoves: Good point! There are no special updates required on our end or for consumers, and this simply brings the following improvements:

- Compose Multiplatform codebase is fully migrated to K2. Please note that native and web klibs can be consumed only with Kotlin 2.1.0 or newer. Also, due to underlying changes in the compiler plugin, it's better to recompile libraries against the new version. 
- The iOS version is now stable 🤩

JayShortway:
@skydoves Do you know what the new minimum Kotlin version is for consumers of the library? If not, would you be able to figure that out? 😄🙏

tonidero (approved):
Looks great! I think we can merge this! Lmk if you need help with the conflicts @skydoves! Maybe we can try to see if we're dropping support for some kotlin version and update the README...

skydoves:
Hey @JayShortway, I’m still figuring it out, but it’s been tough to find any definitive guidance so far. From what I’ve read in the compatibility docs, it looks like customers will need to use at least Kotlin 2.0 since it relies on the K2 compiler.

https://kotlinlang.org/docs/compatibility-guide-21.html#deprecate-freezing-related-apis-in-kotlin-native

tonidero (approved):
:shipit:

Issue #358, opened 2025-03-17
iPad Paywall Not Displaying Properly Until Device is Rotated


hellosagar:
- [ ] I have updated Purchases SDK to the latest version
- [ ] I have read
  the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [ ] I have searched the [Community](https://community.revenuecat.com)
- [ ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ ] I have searched
  for [existing Github issues](https://github.com/RevenueCat/purchases-kmp/issues)

**Describe the bug**
On iPads, the RevenueCat paywall does not always appear correctly when presented but only displayed correctly after the device is rotated

https://github.com/user-attachments/assets/3be5d2fb-9cb1-4ed3-9d21-7df230ab10b8

1. Environment
    1. Platform: Compose Multiplatform
    2. SDK version: 
    3. AGP: 8.5.2
    3. Kotlin: 2.1.0
    3. OS version: Sonoma 14.6.1 (23G93)
    4. IDE (e.g. Android Studio, Xcode, Fleet): Android
    5. IDE version: Android Studio Meerkat | 2024.3.1
    6. Device and/or emulator/simulator:
        - [ ] Device
        - [ ] Emulator/simulator
    7. Environment:
        - [ ] Closed testing / Sandbox
        - [ ] TestFlight
        - [ ] Production
    8. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to
   have context, eg. stackoverflow, etc.)

**Additional context**

hellosagar:
FYI, I've posed the issue here as well - https://community.revenuecat.com/dashboard-tools-52/ipad-paywall-not-displaying-properly-until-device-is-rotated-358-6091

So, Do I need to close one or the other?

JayShortway:
Thanks for reporting this! At first glance it does seem unique to purchases-kmp, as I cannot reproduce it in purchases-ios. So it might be better to discuss here in this repo instead of in the RC Community. 

- Are you by any chance using an animation or transition when showing the paywall? If so, could you try to show it without one? This is just to check if the transition influences this behavior. 
[...]

Issue #287, opened 2024-12-31
Empty paywall after updating Compose Multiplatform to 1.7.3


MrJohn10:
When displaying paywall after updating Compose Multiplatform from 1.6.11 to 1.7.3 the paywall only displays an image. No text/buttons are being displayed.

purchases-kmp = "1.3.6+13.13.1"
compose-plugin = "1.7.3"
kotlin = "2.0.0"

JayShortway:
Hi, thanks for reporting this! Some questions to help investigate this: 
- Would you be able to share a code snippet of how you're displaying the paywall, including the `PaywallOptions` used?
- Does it work correctly with Compose Multiplatform 1.7.0? 
- Does it work correctly on Android?
  
Thanks in advance!

MrJohn10:
`@Composable
fun PaywallDialog(
    displayDismissButton: Boolean = true,
    onDismiss: () -> Unit,
) {
    Paywall(PaywallOptions(dismissRequest = onDismiss) {shouldDisplayDismissButton = true})
}`

It does work on android.
It does not work on iOS with compose 1.7.0

JayShortway:
Thanks for providing that info. Unfortunately I'm not able to reproduce it on my end. Are you showing the Paywall in an alert dialog somehow? Would you be able to share a minimal reproduction project so I can take a look?

MrJohn10:
I'm showing it as a full screen that I navigate to using `org.jetbrains.androidx.navigation`
I will try to reproduce in a sample project next week.

MrJohn10:
It has to be something with the size of the screen. When testing on iPhone 16 pro, only image is visible, but after applying padding to the paywall, everything is correct.

```
    Box(Modifier.fillMaxSize().background(Color.Red).padding(64.dp)) {
        Paywall(options)
    }
```

Tested on the 1. paywall (Jaguar).

When testing on iPhone 8 Plus:
 - image and main content is visible
 - after adding padding everything is visible
 
 Screenshots from iPhone 16 Pro:
 
![Simulator Screenshot - iPhone 16 Pro - 2025-01-04 at 15 08 17](https://github.com/user-attachments/assets/35e92b48-f7fa-45b2-a9d1-b946faf885c2)

 
![simulator_screenshot_2856A3C6-41D1-4E7E-BB89-F92BFEC469BA](https://github.com/user-attachments/assets/24a0eb4c-ac54-4659-8be6-bffe1ce11cda)

[...]

PR #427, opened 2025-06-30
Fixes Paywall and Customer Center transition animations on iOS

### LINKED ISSUE #398: The paywall on iOS is displayed outside the horizontal bounds, and only the first horizontal half is visible
- [X] I have read
- [X] I have searched the [Community](https://community.revenuecat.com)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched

**Describe the bug**
The paywall on iOS is displayed outside the horizontal bounds, and only the first horizontal half is visible. Sometimes it displays correctly only when displaying the paywall for the first time after the app launches; otherwise, it always looks too wide. On Android, it displays correctly.

1. Environment
    1. Platform: iOS
    2. SDK version: 1.7.8+13.32.0
    3. OS version: iOS 18.5
    4. IDE (e.g. Android Studio, Xcode, Fleet): Android Studio and Xcode
    5. IDE version: AS 2024.3.2, XC 16.3 (16E140)
        - [X] Device: iPhone 14 Pro Max, iPhone 11
    7. Environment:
        - [X] Closed testing / Sandbox
        - [X] TestFlight
        - [X] Production
    8. How widespread is the issue. Percentage of devices affected. 100%
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue N/A
     Open the app, launch the Paywall. Expected: Paywall should display correctly, the same as on Android. Actual: Paywall is displayed out of display bounds.
5. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to   have context, eg. stackoverflow, etc.)
Check the attached screenshots.

Android Paywall
Image

iOS Paywall
Image

JayShortway:
## Bug
There's a long standing bug where UIKit views are not rendered properly when they use position-based transition animations (e.g. slide in). This affects both Paywalls and Customer Center.

## Fix
This PR fixes that by updating the Compose constraints on the UIKit side using `NSLayoutConstraint`, in a new `layoutViewController()` modifier. 

| Before | After |
|-------|------|
| https://github.com/user-attachments/assets/01117e6b-a705-4bc1-8961-043ad0270fae |  https://github.com/user-attachments/assets/9c868d15-2561-46ea-ab77-33fd7ce4c730 |
| https://github.com/user-attachments/assets/4fa36f31-2348-43d2-9729-3efb2ede1e35 | https://github.com/user-attachments/assets/b5db47ad-262b-42c8-83ca-4d4ea66f2c5e |

Fixes #398, #358, #287


revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/ConstrainingViewController.kt
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private fun CValue<CGRect>.satisfy(constraints: Constraints): CValue<CGRect> = useContents {
+        CGRectMake(
+            x = origin.x,
+            y = origin.y,
+            width = constraints.fixedWidthPt ?: size.width,
+            height = constraints.fixedHeightPt ?: size.height,
+        )
+    }
```
JayShortway: This is the fix.

ajpallares: 🤔 I'm not really sure `viewWillLayoutSubviews()` is the best place for this. Could this potentially cause an infinite loop? (like `viewWillLayoutSubviews()` makes the view frame change, then this view frame change triggers a new `viewWillLayoutSubviews()` call, and so on). I've seen some infinite loops of this kind with with `layoutSubviews()` (not that I remember any with `viewWillLayoutSubviews()` tbh).

Now, since this fully seems to depend on the `constraints` variable, why not use iOS `NSLayoutConstraints` instead? If I'm understanding the code correctly, we only care about the fixedWidth and fixedHeight values of the constraints object. Don't know if this is possible, but could we react when `var constraints` changes so that we:
1. Invalidate the previously-stored `NSLayoutConstrains` array
2. Create, activate and store a new `NSLayoutConstraints` array (based on `constraints.fixedWidthPt` and `constraints.fixedHeightPt`, i.e. if `constraints.fixedWidthPt` is `null`, do not add a fixed width constraint, and similar if `constraints.fixedHeightPt` is `null`)

My experience in the past has been that working with frames in `viewWillLayoutSubviews` and similar methods is flaky, whereas using `NSLayoutConstraints` has been very reliable.

Also, if you take this approach, you make sure you execute the updateNSLayoutConstraints logic in `viewDidLoad()` as well, as it might happen that the `constraints` var is set before the `view` is loaded.

WDYT?

ajpallares (commented):
Wow, nice work fixing these issues! I can mostly add value regarding the iOS code. Please, see my comment with a proposal to use `NSLayoutConstraints` instead. Let me know what you think!

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/ConstrainingViewController.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        view.setFrame(view.frame.satisfy(currentConstraints))
+    }
+
+    private fun CValue<CGRect>.satisfy(constraints: Constraints): CValue<CGRect> = useContents {
+        CGRectMake(
+            x = origin.x,
+            y = origin.y,
+            width = constraints.fixedWidthPt ?: size.width,
+            height = constraints.fixedHeightPt ?: size.height,
```
tonidero: Just in case, have you tested the fix on footer paywalls both loaded, and the transition from loading to loaded?

tonidero (commented):
No other real comments aside from existing ones. But nice work on this!!

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/ConstrainingViewController.kt
JayShortway: Amazing, thank you! This is exactly why I requested your review specifically 😄 Will have a look.

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/ConstrainingViewController.kt
JayShortway: Great callout! I added a toggle to the sample app to test this more easily. 

It still works: 

https://github.com/user-attachments/assets/1ff70342-38ba-4907-a302-cada62adb98a

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/ConstrainingViewController.kt
JayShortway: I moved all of the layout logic into the new custom `layoutViewController()` modifier, and the `LayoutViewController` Modifier Node. This greatly simplified both `UIKitPaywall` and `UIKitCustomerCenter`.

JayShortway: @ajpallares Thanks again for all your help! 🙌 🫶

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/modifier/layoutViewController.kt
```diff
@@ -0,0 +1,146 @@
...
+        constraints: Constraints,
+    ): MeasureResult {
+        // Check if we are being asked to wrap our own content height. If so, we will use the
+        // measurement done by UIKit.
+        val constraintsToUse = if (!constraints.hasFixedHeight)
+            intrinsicContentHeightProvider()
+                .roundToPx()
+                .coerceAtMost(constraints.maxHeight)
+                .let { height -> constraints.copy(minHeight = height, maxHeight = height) }
+        else constraints
```
ajpallares: Would it ever be needed to do the same for the width? I'm mostly thinking on iPad, where a paywall could potentially not cover the whole screen width.

ajpallares (approved):
We did end up with a very clean solution. Great job! 👏

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/modifier/layoutViewController.kt
JayShortway: Good point! Can't hurt. Added here: https://github.com/RevenueCat/purchases-kmp/pull/427/commits/4e5cbff19f0c5c0a818c456bddadac581c0c06a5.

PR #469, opened 2025-08-12
Migrates to `kotlin.time.Instant` from the stdlib

### LINKED ISSUE #468: Update KotlinX Date/Time to 0.7.1
I'm having an issue while trying to push the latest version of my app, because my app's latest KotlinX Date/Time version do not align with yours. There have been some [changes](https://github.com/Kotlin/kotlinx-datetime/releases/tag/v0.7.1) in the meantime, so I would suggest modifying it accordingly.

https://github.com/RevenueCat/purchases-kmp/blob/main/gradle/libs.versions.toml#L22

```
Missing class kotlinx.datetime.Instant$Companion (referenced from: java.util.Map com.revenuecat.purchases.kmp.datetime.CustomerInfoKt.getAllExpirationInstants(com.revenuecat.purchases.kmp.models.CustomerInfo) and 11 other contexts)
Missing class kotlinx.datetime.Instant (referenced from: java.util.Map com.revenuecat.purchases.kmp.datetime.CustomerInfoKt.getAllExpirationInstants(com.revenuecat.purchases.kmp.models.CustomerInfo) and 11 other contexts)
```

JayShortway:
## Description

The `Instant` (and `Clock`) types [are moving](https://github.com/Kotlin/kotlinx-datetime?tab=readme-ov-file#deprecation-of-instant) from the kotlinx-datetime library to Kotlin's stdlib. They're available since Kotlin 2.1.20. Concretely this means that `kotlinx.datetime.Instant` is deprecated and replaced by `kotlin.time.Instant`. This PR makes use of the new type, and removes the kotlinx-datetime dependency (because we don't use anything else). 

This is technically a breaking change, but the kotlinx-datetime library is still experimental in its entirety. This will only affect devs using that experimental library already. 

The alternative is to bump the version of the kotlinx-datetime library to 0.7.1, which is a compatibility release still containing the deprecated types, but that's just delays the inevitable. Happy to hear thoughts.

Closes #468

### DIFF
--- a/datetime/src/commonMain/kotlin/com/revenuecat/purchases/kmp/datetime/CustomerInfo.kt
+++ b/datetime/src/commonMain/kotlin/com/revenuecat/purchases/kmp/datetime/CustomerInfo.kt
@@ -2,13 +2,15 @@ package com.revenuecat.purchases.kmp.datetime
 
 import com.revenuecat.purchases.kmp.Purchases
 import com.revenuecat.purchases.kmp.models.CustomerInfo
-import kotlinx.datetime.Instant
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
 
 /**
  * Map of productIds to expiration instants.
  *  * For Google subscriptions, productIds are `subscriptionId:basePlanId`.
  *  * For Amazon subscriptions, productsIds are `termSku`.
  */
+@OptIn(ExperimentalTime::class)
 public val CustomerInfo.allExpirationInstants: Map<String, Instant?>
     get() = allExpirationDateMillis.mapValues { (_, millis) ->
         millis?.let { Instant.fromEpochMilliseconds(it) }
@@ -20,6 +22,7 @@ public val CustomerInfo.allExpirationInstants: Map<String, Instant?>
  *  * For Google and Amazon INAPPs, productsIds are simply `productId`.
  *  * For Amazon subscriptions, productsIds are `termSku`.
  */
+@OptIn(ExperimentalTime::class)
 public val CustomerInfo.allPurchaseInstants: Map<String, Instant?>
     get() = allPurchaseDateMillis.mapValues { (_, millis) ->
         millis?.let { Instant.fromEpochMilliseconds(it) }
@@ -28,12 +31,14 @@ public val CustomerInfo.allPurchaseInstants: Map<String, Instant?>
 /**
  * The instant this user was first seen in RevenueCat.
  */
+@OptIn(ExperimentalTime::class)
 public val CustomerInfo.firstSeenInstant: Instant
     get() = Instant.fromEpochMilliseconds(firstSeenMillis)
 
 /**
  * The latest expiration instant of all purchased productIds.
  */
+@OptIn(ExperimentalTime::class)
 public val CustomerInfo.latestExpirationInstant: Instant?
     get() = latestExpirationDateMillis?.let { Instant.fromEpochMilliseconds(it) }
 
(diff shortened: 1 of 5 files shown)


datetime/src/commonMain/kotlin/com/revenuecat/purchases/kmp/datetime/CustomerInfo.kt
```diff
@@ -2,13 +2,15 @@ package com.revenuecat.purchases.kmp.datetime
...
-import kotlinx.datetime.Instant
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
 
 /**
  * Map of productIds to expiration instants.
  *  * For Google subscriptions, productIds are `subscriptionId:basePlanId`.
  *  * For Amazon subscriptions, productsIds are `termSku`.
  */
+@OptIn(ExperimentalTime::class)
```
JayShortway, line 13: Maybe it's better to propagate these annotations?

datetime/api/datetime.api
```diff
@@ -1,21 +1,21 @@
 public final class com/revenuecat/purchases/kmp/datetime/CustomerInfoKt {
 	public static final fun getAllExpirationInstants (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;)Ljava/util/Map;
 	public static final fun getAllPurchaseInstants (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;)Ljava/util/Map;
-	public static final fun getFirstSeenInstant (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;)Lkotlinx/datetime/Instant;
-	public static final fun getLatestExpirationInstant (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;)Lkotlinx/datetime/Instant;
-	public static final fun getOriginalPurchaseInstant (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;)Lkotlinx/datetime/Instant;
-	public static final fun getRequestInstant (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;)Lkotlinx/datetime/Instant;
+	public static final fun getFirstSeenInstant (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;)Lkotlin/time/Instant;
```
tonidero, line 4: Hmm as you said this is a breaking change. I think it would be ok to do this, as long as devs had to previously OptIn into an experimental flag to be able to use them. If not, I would consider holding using the compatibility library version if possible.

datetime/src/commonMain/kotlin/com/revenuecat/purchases/kmp/datetime/CustomerInfo.kt
tonidero: If we're using types that are experimental, I think that makes sense to me!

datetime/api/datetime.api
tonidero: My main question is that before it seems we were able to use these types in the API tests without opting in to the experimental flag.

JayShortway: It seems that there's no way around that. Even updating to 0.7.1 and using `kotlinx.datetime.Instant` (the deprecated type), we're forced to add the experimental flag.
  
image

tonidero: Ugh :( So... I guess we're forced to do a breaking change... I really would love to avoid a major but wondering if we should do it, even if it's due to a dependency...

datetime/src/commonMain/kotlin/com/revenuecat/purchases/kmp/datetime/CustomerInfo.kt
skydoves: Agreed to propagate these annotations.

skydoves (approved):
Overall looks great to me!

PR #513, opened 2025-09-29
[v1] Prepare for next release

ajpallares:
This PR makes the changes needed for the next release of v1.

* Bump [fastlane-plugin-revenuecat_internal](https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal) from `7d97553 ` to `1593f78`. See full diff in <a href="https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/compare/7d97553e9c5baabcd18286f03d8034797a27dd64...1593f78d0b9b24b48238337666183e3ba82f848e">compare view</a>.
* Pass current version in automatic bump.
* [CI] `bump` pipeline action won't trigger `on-release-branch` workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -474,6 +474,8 @@ workflows:
         - matches: { pattern: "^release/.+$", value: << pipeline.git.branch >> }
         - not:
             equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+        - not:
+            equal: [ bump, << pipeline.parameters.action >> ]
     jobs:
       - detekt
       - build-libraries-android
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal
-  revision: 7d97553e9c5baabcd18286f03d8034797a27dd64
+  revision: 1593f78d0b9b24b48238337666183e3ba82f848e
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -474,6 +474,10 @@ workflows:
         - matches: { pattern: "^release/.+$", value: << pipeline.git.branch >> }
         - not:
             equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+        # Only run this workflow when the action is "build" (the default), 
+        # so it doesn't run for other actions like "bump" on a release branch,
+        # which should only trigger the release train workflow but not this one.
+        - equal: [ build, << pipeline.parameters.action >> ]
```
ajpallares, line 481: @JayShortway do you think this is clear enough? It's a bit tricky to explain without extending a lot... 😅

JayShortway: Maybe this? 🤷 It's not the most succinct, but this would have made me understand it right away I think haha.

```suggestion
        # Only run this workflow when the action is "build" (the default), 
        # so it doesn't run for other actions like "bump" on a release branch,
        # which should only open the release PR but not deploy it.
        # (Deployment happens after the hold job is approved on the release PR.)
        - equal: [ build, << pipeline.parameters.action >> ] 
```

ajpallares: That's better. Thank you for the suggestion! 🫶

PR #514, opened 2025-09-29
Moves Instant properties to `:purchases-kmp-core`

JayShortway:
## Description
Round 2 of https://github.com/RevenueCat/purchases-kmp/pull/469, this time without breaking changes!

- Deprecates all `kotlinx.datetime.Instant` properties in `:purchases-kmp-datetime`, and
- Adds them as `kotlin.time.Instant` properties in `:purchases-kmp-core` (via `:purchases-kmp-models`).

Also updates the patch version of Kotlin to 2.1.21. We need 2.1.20 for `kotlin.time.Instant`. [2.1.21 gives us Xcode 16.3 compatibility](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-compatibility-guide.html#version-compatibility), up from 16.0.  

Closes https://github.com/RevenueCat/purchases-kmp/issues/468


gradle/libs.versions.toml
```diff
@@ -6,7 +6,7 @@ android-targetSdk = "35"
 ios-deploymentTarget-core = "13.0"
 ios-deploymentTarget-ui = "15.0"
 java = "1.8"
-kotlin = "2.1.10"
+kotlin = "2.1.21"
```
tonidero, line 9: Nitpick. I don't see anything that could cause issues in the changes here, but would it be worth extracting this to a separate PR? Mostly so it shows in the changelog separately and it's clearer when the bump happened.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomerInfo.kt
```diff
@@ -88,4 +91,66 @@ public class CustomerInfo(
...
+) {
+
+
+    /**
+     * Map of productIds to expiration dates.
+     *  * For Google subscriptions, productIds are `subscriptionId:basePlanId`.
+     *  * For Amazon subscriptions, productsIds are `termSku`.
+     */
+    @ExperimentalTime
+    public val allExpirationDates: Map<String, Instant?> by lazy {
```
tonidero, line 103: I see we used the `instant` naming for these in the previous module, like `allExpirationInstants`, probably to avoid the naming conflict which would cause a breaking change... I kinda liked that the previous name had the type in it, but I don't see a better alternative, and I think this is probably ok, specially considering it's probably the most common way to represent these now 👍

tonidero (approved):
Looks great! Thanks for handling this 🎉

gradle/libs.versions.toml
JayShortway: That's a great idea! Separate PR is here: https://github.com/RevenueCat/purchases-kmp/pull/515.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomerInfo.kt
JayShortway: Yea, I wanted to avoid a naming conflict indeed. If they're named the same as the deprecated properties, the new properties shadow the deprecated ones. This means that when developers update purchases-kmp to the version this PR gets released in, they would get build errors because of `@ExperimentalTime` (because they are unknowingly using the new properties).

PR #521, opened 2025-09-30
Adds purchases-ios as git submodule

JayShortway:
## Motivation
We want to improve the developer experience (both for app developers and ourselves) of purchases-kmp, by removing the need to manually declare the PHC dependency in the iOS app. We will instead embed purchases-ios statically in the shipped `.klib`.

## Description
We'll do this incrementally. This is the first of many steps. It just adds purchases-ios as a submodule, but it's still unused. Later steps will make sure the purchases-ios source code gets compiled as part of the build, and included in the `.klib`.

- Adds purchases-ios as git submodule.
- Adds a Renovate config to keep it updated.

I purposefully did not set purchases-ios to the latest version, because I want to see if the Renovate config works. Renovate is kinda hard to test until it's all on `main`. We won't be using the submodule for a while yet, so we have some time to figure this out.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "upstream/purchases-ios"]
 	path = upstream/purchases-ios
 	url = git@github.com:RevenueCat/purchases-ios.git
+	branch = 5.39.3
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "upstream/purchases-ios"]
+	path = upstream/purchases-ios
+	url = git@github.com:RevenueCat/purchases-ios.git
--- a/renovate.json
+++ b/renovate.json
@@ -1,4 +1,8 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["github>RevenueCat/renovate"]
+  "extends": ["github>RevenueCat/renovate"],
+  "git-submodules": {
+    "enabled": true,
+    "versioning": "semver"
+  }
 }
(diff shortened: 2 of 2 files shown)


.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "upstream/purchases-ios"]
+	path = upstream/purchases-ios
+	url = git@github.com:RevenueCat/purchases-ios.git
+	branch = 5.39.3
```
tonidero, line 4: Hmm do we need to change this/update it when updating the PHC version? Though I guess renovate will be taking care of that :), this is not the current latest version of purchases-ios.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "upstream/purchases-ios"]
```
tonidero, line 1: I guess we're still not initializing submodules in CircleCI on each job, but I guess that will be coming in future PRs :)

ajpallares (approved):

.gitmodules
JayShortway: Yes yes 😄 

>I purposefully did not set purchases-ios to the latest version, because I want to see if the Renovate config works. Renovate is kinda hard to test until it's all on main. We won't be using the submodule for a while yet, so we have some time to figure this out.

.gitmodules
JayShortway: Ah good point! But yes, that can come in future PRs, when this starts to be used. For now, this is only being developed locally anyway.

.gitmodules
tonidero: I should learn to read 🤦 😂

JayShortway: Hahah 🫶

PR #553, opened 2025-11-03
Makes `PurchasesApplicationProvider` public so you can configure RevenueCat in AndroidX Startup `Initializer`s

### LINKED ISSUE #551: Purchases has no reference to the Application. Please make sure you have not removed the androidx.startup.InitializationProvider from your AndroidManifest.xml
- [ x] I have updated Purchases SDK to the latest version
- [ x] I have read
- [ x] I have searched the [Community](https://community.revenuecat.com)
- [ x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ x] I have searched

**Describe the bug**
This is an error initialising RevenueCat on Android. I’m not removing the initializer. My existing manifest has:

```
<provider
    android:name="androidx.startup.InitializationProvider"
    android:authorities="${applicationId}.androidx-startup"
    android:exported="false"
    tools:node="merge">
    <meta-data
        android:name="com.foo.BarAppInitializer"
        android:value="androidx.startup" />
</provider>
```

I also couldn’t find a workaround to inject the application context myself, because the `AndroidProvider` is internal visibility.

1. Environment
    1. Platform: Android
    2. SDK version: 2.2.4+17.12.0
    3. OS version: 16 (36)
    4. IDE (e.g. Android Studio, Xcode, Fleet): Android Studio
    5. IDE version: 2025.1.4
        - [ x] Device
        - [ ] Emulator/simulator
    7. Environment:
        - [ x] Closed testing / Sandbox
        - [ ] TestFlight
        - [ ] Production
    8. How widespread is the issue. Percentage of devices affected. 100%

Purchases has no reference to the Application. Please make sure you have not removed the androidx.startup.InitializationProvider from your AndroidManifest.xml. If you need to remove specific initializers, such as androidx.work.WorkManagerInitializer, do so as follows:
    
```
    <provider
        android:name="androidx.startup.InitializationProvider"
        android:exported="false"
        android:authorities="${applicationId}.androidx-startup"
        tools:node="merge">
    
        <meta-data
            android:name="androidx.work.WorkManagerInitializer"
            android:value="androidx.startup"
            tools:node="remove" />
    
    </provider>
```

Just run the app with revenuecat configured

4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to   have context, eg. stackoverflow, etc.)
```    
    	at com.revenuecat.purchases.kmp.di.AndroidProviderKt.requireApplication(AndroidProvider.kt:69)
    	at com.revenuecat.purchases.kmp.Purchases$Companion.configure(Purchases.android.kt:116)
```
**Additional context**

JayShortway:
## Description
Before this PR it was not possible to call `Purchases.configure()` in an [AndroidX Startup](https://developer.android.com/topic/libraries/app-startup#lint-checks) `Initializer`, because it's impossible to guarantee that RevenueCat's `PurchasesApplicationProvider` runs before your own.  

This PR makes `PurchasesApplicationProvider` public, so you can add it to your own `Initializer`'s [dependencies](https://developer.android.com/reference/kotlin/androidx/startup/Initializer#dependencies()), like so: 

```kotlin
override fun dependencies(): List<Class<out Initializer<*>>> = listOf(
    PurchasesApplicationProvider::class.java
)
```

This guarantees `PurchasesApplicationProvider` is called before your own `Initializer`, which means you can now call `Purchases.configure()`. 

Closes #551.

### DIFF
--- a/core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/PurchasesApplicationProvider.kt
+++ b/core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/PurchasesApplicationProvider.kt
@@ -6,13 +6,13 @@ import androidx.startup.Initializer
 import com.revenuecat.purchases.kmp.di.AndroidProvider
 
 /**
- * This class is only used to capture the [android.app.Application] instance and keep a reference to it. The
+ * This class is only used to capture the [Application] instance and keep a reference to it. The
  * `Application` instance is used for 2 things:
  * 1. To call static Purchases methods that need a Context, such as
  * [canMakePayments][com.revenuecat.purchases.Purchases.canMakePayments].
  * 2. To keep track of the current Activity, which is needed to make purchases.
  *
- * If you need to use RevenueCat in your own [androidx.startup.Initializer], make sure to add
+ * If you need to use RevenueCat in your own [Initializer], make sure to add
  * [PurchasesApplicationProvider] to its dependencies like so:
  *
  * ```kotlin


core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/di/PurchasesInitializer.kt
tonidero: Hmm should we move this to a different package? Since weird to expose a class in a `di` package, which feels more internal... Maybe move it to top-level?

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/di/PurchasesInitializer.kt
```diff
@@ -10,8 +10,17 @@ import androidx.startup.Initializer
...
+ * to its dependencies like so:
+ *
+ * ```kotlin
+ * override fun dependencies(): List<Class<out Initializer<*>>> = listOf(
+ *     PurchasesInitializer::class.java
+ * )
+ * ```
  */
-internal class PurchasesInitializer : Initializer<Unit> {
+public class PurchasesInitializer : Initializer<Unit> {
```
tonidero: I'm just wondering if we should rename this to something more specific like `PurchasesActivityTrackingInitializer`... Feels like this would actually initialize `Purchases`, which it doesn't... Not a strong opinion though.

tonidero (approved):
Just a comment about moving the class to a different package and about a possible rename. Looks good otherwise!

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/di/PurchasesInitializer.kt
JayShortway: Good point, done here: https://github.com/RevenueCat/purchases-kmp/pull/553/commits/fa70b7558b50d728fd406dba51411e048c1087ae

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/di/PurchasesInitializer.kt
JayShortway: Very good point! As discussed, renamed to `PurchasesApplicationProvider` here: https://github.com/RevenueCat/purchases-kmp/pull/553/commits/7bdba1f3f18b9acd17a8011c7a5e9b8ed7085941.

core/build.gradle.kts
```diff
@@ -12,8 +12,8 @@ kotlin {
             api(projects.models)
         }
         androidMain.dependencies {
+            api(libs.androidx.startup)
```
tonidero, line 15: I guess we need to expose this dependency since our class inherits from Initializer right? 🥲

JayShortway: Yea.. technically `androidx.startup.Initializer` is now part of our API 🙁

PR #624, opened 2026-01-05
Merges Gradle convention plugins

JayShortway:
## Description

We had 2 Gradle convention plugins: `InternalLibraryConventionPlugin` and `PublicLibraryConventionPlugin`. The only difference is that the latter generates public doc pages.

This PR merges the 2 plugins into 1, and introduces a so-called extension to configure whether docs should be generated. We can now do this in `build.gradle.kts` files:
```kotlin
revenueCat {
    dokka = true // defaults to false
}
```


build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/plugin/RevenueCatLibraryConventionPlugin.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+class RevenueCatLibraryConventionPlugin : Plugin<Project> {
+
+    override fun apply(target: Project) = with(target) {
+        val extension = extensions.create("revenueCat", RevenueCatExtension::class.java)
+
+        with(pluginManager) {
+            apply("org.jetbrains.kotlin.multiplatform")
+            apply("com.android.library")
+            apply("dev.adamko.dokkatoo-html")
```
rickvdl, line 20: Couldn't we do the same here as we did previously?
```
 if (dokka) apply("dev.adamko.dokkatoo-html")
```

Or is this information (now `extension.dokka`) not available at this point yet?

JayShortway: Yea that's exactly it. Plugins are applied before the extension is evaluated, so we have to take a different approach.

rickvdl: Makes sense, thanks for clarifying!

PR #625, opened 2026-01-06
Allows declaring `swiftPackage` dependencies

JayShortway:
## Description

This adds functionality to our Gradle plugin that allows us to depend on a Swift package like so:

```kotlin
kotlin {
    sourceSets {
        iosMain.dependencies {
            swiftPackage(
                path = rootProject.file("upstream/purchases-ios"),
                target = "RevenueCat",
                packageName = "swiftPMImport.com.revenuecat.purchases.kn.core",
            )
        }
    }
}
```

This is not yet used anywhere. This PR just adds the capability. 

## Motivation
I'm working on improving the integration experience of purchases-kmp, by avoiding the need for app developers to add the purchases-hybrid-common dependency manually as this adds friction and prevents properly testing Kotlin/Native modules that depend on purchases-kmp. Instead, we will be embedding purchases-ios directly in purchases-kmp, handling all complexity for the app developers. This PR introduces all the build logic required for that.

## How it works
Most orchestration happens in `ConfigureSwiftDependencies`:
1. All calls to `swiftPackage()` are collected.
2. Each Swift source directory is resolved.
3. A `GenerateDefFileTask` is registered for each target, to generate [a def file](https://kotlinlang.org/docs/native-definition-file.html) for Kotlin/Native interop.
4. A `SwiftBuildTask` is registered for each target, which will compile the Swift code.
5. A cinterop configuration is created for each Kotlin target that depends on any Swift target. This will generate Kotlin bindings for the Objective-C headers in our Swift target.
6. The cinterop task is set to depend on the `GenerateDefFileTask` and `SwiftBuildTask`, to ensure the correct order of execution.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,6 +5,7 @@ org.gradle.configuration-cache=true
 
 #Kotlin
 kotlin.code.style=official
+kotlin.mpp.enableCInteropCommonization=true
 
 #Android
 android.useAndroidX=true
--- a/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/plugin/RevenueCatLibraryConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/plugin/RevenueCatLibraryConventionPlugin.kt
@@ -2,6 +2,7 @@ package com.revenuecat.purchases.kmp.buildlogic.plugin
 
 import com.revenuecat.purchases.kmp.buildlogic.convention.configureAndroid
 import com.revenuecat.purchases.kmp.buildlogic.convention.configureKotlin
+import com.revenuecat.purchases.kmp.buildlogic.swift.configureSwiftDependencies
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 
@@ -24,6 +25,7 @@ class RevenueCatLibraryConventionPlugin : Plugin<Project> {
 
         configureKotlin()
         configureAndroid()
+        configureSwiftDependencies()
 
         // Disable dokka if needed.
         afterEvaluate {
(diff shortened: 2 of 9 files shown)


build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/test/revenueCatLibraryPluginTest.kt
```diff
@@ -0,0 +1,242 @@
...
+import org.gradle.testkit.runner.GradleRunner
+import org.intellij.lang.annotations.Language
+import java.io.File
+import kotlin.io.path.createTempDirectory
+
+
+/**
+ * Run an end-to-end test of the RevenueCatLibraryConvention plugin.
+ */
+fun revenueCatLibraryPluginTest(
```
JayShortway: This is a little testing DSL that allows us to create a Gradle project on the fly to run tests on. It supports things like:
```kotlin
revenueCatLibraryPluginTest {
        val pkg = addSwiftPackage(kotlinSourceSet = "iosMain") {
            writeTestClass(relativePath = "TestClass.swift")
        }
        runBuild(pkg.cinteropTaskName)
        pkg.target.readSourceFile("TestClass.swift")
        pkg.target.writeSourceFile("TestClass.swift", "$original\n// comment\n")
}
```

It's used in `SwiftCinteropUpToDateTests.kt` below.

gradle.properties
```diff
@@ -5,6 +5,7 @@ org.gradle.configuration-cache=true
 
 #Kotlin
 kotlin.code.style=official
+kotlin.mpp.enableCInteropCommonization=true
```
JayShortway, line 8: This is needed to generate Kotlin bindings in `iosMain`, which is a "common" source set. Without this setting, Kotlin bindings are only generated in the underlying source sets: `iosX64Main`, `iosArm64Main` and `iosSimulatorArm64`.

PR #626, opened 2026-01-06
Adds tests for `swiftPackage`

JayShortway:
## Description
This adds some tests for the `swiftPackage` functionality introduced by https://github.com/RevenueCat/purchases-kmp/pull/625. It mainly checks whether the Swift code is properly recompiled when it is changed, to avoid running outdated Swift code. This would be confusing for us at best, and cause us to ship the wrong code at worst.

### DIFF
--- a/build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/test/revenueCatLibraryPluginTest.kt
+++ b/build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/test/revenueCatLibraryPluginTest.kt
@@ -189,17 +189,13 @@ private data class SwiftPackageConfig(
 )
 
 private fun setupBaseProject(projectDir: File) {
+    val mainProjectDir = File(System.getProperty("mainProjectDir")
+        ?: error("rootProjectDir system property not set."))
+    val mainProjectVersionCatalog = mainProjectDir.resolve("gradle/libs.versions.toml")
+    
     projectDir.resolve("gradle").apply {
         mkdirs()
-        resolve("libs.versions.toml").writeText(
-            // language=toml
-            """
-            [versions]
-            android-compileSdk = "35"
-            android-minSdk = "21"
-            java = "1.8"
-            """.trimIndent()
-        )
+        resolve("libs.versions.toml").writeText(mainProjectVersionCatalog.readText())
     }
 
     projectDir.resolve("settings.gradle.kts").writeText(
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -37,6 +37,9 @@ tasks.test {
         events("passed", "skipped", "failed")
         showStandardStreams = true
     }
+    // Provide the main project root directory to tests (build-logic is a composite build,
+    // so rootProject.rootDir refers to build-logic, not the main project)
+    systemProperty("mainProjectDir", rootProject.rootDir.parentFile.absolutePath)
 }
 
 gradlePlugin {


build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/test/revenueCatLibraryPluginTest.kt
```diff
@@ -0,0 +1,242 @@
...
+import org.gradle.testkit.runner.GradleRunner
+import org.intellij.lang.annotations.Language
+import java.io.File
+import kotlin.io.path.createTempDirectory
+
+
+/**
+ * Run an end-to-end test of the RevenueCatLibraryConvention plugin.
+ */
+fun revenueCatLibraryPluginTest(
```
JayShortway, line 13: This is a little testing DSL that allows us to create a Gradle project on the fly to run tests on. It supports things like:
```kotlin
revenueCatLibraryPluginTest {
        val pkg = addSwiftPackage(kotlinSourceSet = "iosMain") {
            writeTestClass(relativePath = "TestClass.swift")
        }
        runBuild(pkg.cinteropTaskName)
        pkg.target.readSourceFile("TestClass.swift")
        pkg.target.writeSourceFile("TestClass.swift", "$original\n// comment\n")
}
```

It's used in `SwiftCinteropUpToDateTests.kt` below.

.circleci/config.yml
```diff
@@ -345,6 +345,20 @@ jobs:
       - save-kotlin-native-compiler-to-cache
       - save-incremental-gradle-build-to-workspace
 
+  unit-tests-build-logic:
+    executor: xcode16
+    steps:
+      - checkout
+      - restore-gradle-user-home-directory-from-cache
```
rickvdl, line 352: Not sure if I understand the workings of this correctly, but could this cause build cache to be shared across runs? E.g. the swift change you test later to be already present in the build cache of the next run?

rickvdl (commented):
Great addition to the test suite, makes sense! 👌 I'm not sure if I even enough context on the specifics yet, so it might be good if someone else also takes a look.

build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/test/revenueCatLibraryPluginTest.kt
```diff
@@ -0,0 +1,242 @@
...
+    } finally {
+        projectDir.deleteRecursively()
+    }
+}
+
+
+/**
+ * Entry-point for our test DSL.
+ */
+class RevenueCatLibraryPluginTestContext(
```
tonidero, line 29: Just wondering if we want the name to reflect this is only for iOS right now (like `RevenueCatSwiftLibraryPluginTestContext`), or leave it generic in case we need to add something else in the future... NABD though, happy to leave as is.

build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/test/revenueCatLibraryPluginTest.kt
```diff
@@ -0,0 +1,242 @@
...
+)
+
+private fun setupBaseProject(projectDir: File) {
+    projectDir.resolve("gradle").apply {
+        mkdirs()
+        resolve("libs.versions.toml").writeText(
+            // language=toml
+            """
+            [versions]
+            android-compileSdk = "35"
```
tonidero: Hmm would be great if we could use the toml properties, to avoid hardcoding them here... Maybe we could just "copy" our own toml file for this? Might be a lot of extra unnecessary declarations, but as long as they are not used later, it would be ok?

.circleci/config.yml
JayShortway: Good callout! The `~/.gradle` directory mostly contains cached dependencies, not intermediate build artifacts. Also, the Swift files (and the entire Gradle project under test) are saved in a temp directory. So I don't expect this to interfere between test runs. We can keep an eye out for it though!

build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/test/revenueCatLibraryPluginTest.kt
JayShortway: Our Gradle plugin is called `RevenueCatLibraryConventionPlugin`, and this was indeed intended to be a generic DSL to test our entire plugin, not necessarily only the Swift-related logic. (`runBuild()` for instance works even without adding Swift files first.)

build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/test/revenueCatLibraryPluginTest.kt
JayShortway: Good one! Implemented here: https://github.com/RevenueCat/purchases-kmp/pull/626/changes/0b5243dc9f704a07abb22040a0c97d04dbb6665d.

Issue #410, opened 2025-06-11
Provide a simple way to detect if we should show 'manage subscription' link in the app


MrJohn10:
We should show 'manage subscription' link in the app not when user has active subscription, but when user has any subscription that is not expired/revoked. For example we should show it when subscription is on hold due to a billing issue or is paused (on Android).

I tested the case of billing issue on iOS in the sandbox environment. In that case:
- EntitlementInfo.willRenew() returns false
- customerInfo.managementUrlString returns null

Is there currently a way to detect if user can be charged for any subscription in the future? I think that should specify if we show them link to manage subscription or not.

MrJohn10:
Any updates on this?

MrJohn10:
I don't think there is a clear way to check if subscription is in the `SUBSCRIPTION_STATE_ON_HOLD` state? Maybe we could just add `playStoreSubscriptionState` that just is the subscription state from google the api?

[Subscription lifecycle](https://developer.android.com/google/play/billing/lifecycle/subscriptions)

JayShortway:
Hi @MrJohn10, is my understanding correct that the ultimate goal is to have a management URL for users in "account hold" state?

PR #633, opened 2026-01-13
Adds richer `SubscriptionInfo` to `CustomerInfo`

JayShortway:
## Description
Brings the `SubscriptionInfo` type we have in purchases-android and purchases-ios to purchases-kmp.

Revamped https://github.com/RevenueCat/purchases-kmp/pull/599. 
Fixes https://github.com/RevenueCat/purchases-kmp/issues/410.

### DIFF
--- a/composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/components/CustomerInfoSection.kt
+++ b/composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/components/CustomerInfoSection.kt
@@ -65,6 +65,10 @@ internal fun CustomerInfoSection(
                         label = "allExpirationDates"
                     )
                     EntitlementInfosRow(infos = state.value.entitlements)
+                    SubscriptionInfoMapRow(
+                        subscriptions = state.value.subscriptionsByProductIdentifier,
+                        label = "subscriptionsByProductIdentifier"
+                    )
                     TransactionsRow(
                         transactions = state.value.nonSubscriptionTransactions,
                         label = "nonSubscriptionTransactions"
--- a/composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/components/SubscriptionInfoMapRow.kt
+++ b/composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/components/SubscriptionInfoMapRow.kt
@@ -1,0 +1,70 @@
+package com.revenuecat.purchases.kmp.sample.components
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import com.revenuecat.purchases.kmp.models.SubscriptionInfo
+import com.revenuecat.purchases.kmp.sample.DefaultPaddingHorizontal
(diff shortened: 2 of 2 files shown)


mappings/src/iosMain/kotlin/com/revenuecat/purchases/kmp/mappings/SubscriptionInfo.ios.kt
```diff
@@ -0,0 +1,54 @@
...
+        unsubscribeDetectedAtMillis = (obj.valueForKey("unsubscribeDetectedAt") as? NSDate)?.toEpochMilliseconds(),
+        billingIssuesDetectedAtMillis = (obj.valueForKey("billingIssuesDetectedAt") as? NSDate)?.toEpochMilliseconds(),
+        gracePeriodExpiresDateMillis = (obj.valueForKey("gracePeriodExpiresDate") as? NSDate)?.toEpochMilliseconds(),
+        ownershipType = (obj.valueForKey("ownershipType") as RCPurchaseOwnershipType).toOwnershipType(),
+        periodType = (obj.valueForKey("periodType") as RCPeriodType).toPeriodType(),
+        refundedAtMillis = (obj.valueForKey("refundedAt") as? NSDate)?.toEpochMilliseconds(),
+        storeTransactionId = obj.valueForKey("storeTransactionId") as? String,
+        autoResumeDateMillis = null, // Not available on iOS
+        price = subscriptionPrice,
+        productPlanIdentifier = null, // Not available on iOS
```
tonidero, line 49: Hmm couldn't this be available in iOS, for subscriptions that come from Android? I just checked and we're not parsing it in iOS though, so I don't think we can add it here yet. We might want to revisit the iOS part at some point though.

mappings/src/iosMain/kotlin/com/revenuecat/purchases/kmp/mappings/SubscriptionInfo.ios.kt
```diff
@@ -0,0 +1,54 @@
...
+                currencyCode = currency,
+                amountMicros = (amount * 1_000_000.0).toLong(),
+                formatted = formatted
+            )
+        } else {
+            null
+        }
+    }
+
+    return SubscriptionInfo(
```
tonidero, line 33: Really will love once we add more integration tests using the test store to these hybrid SDKs, which should allow us to test these mappings as well, fully end to end 🙏

tonidero (approved):
Looking great to me!! Thanks for adding this!

mappings/src/iosMain/kotlin/com/revenuecat/purchases/kmp/mappings/SubscriptionInfo.ios.kt
JayShortway: This will all be gone once we move to the implementation I started [here](https://github.com/RevenueCat/purchases-kmp/pull/625). 😄 

Agreed on end to end tests regardless of course!

mappings/src/iosMain/kotlin/com/revenuecat/purchases/kmp/mappings/SubscriptionInfo.ios.kt
JayShortway: Paging @vegaro for implementation details 😄

vegaro: Oh that's a good point. Yeah it's not in purchases-ios but can definitely be useful there for those Android subcriptions

PR #640, opened 2026-01-20
Expose ad tracking APIs in KMP SDK for Android

polmiro:
### Checklist

### Motivation

The RevenueCat KMP SDK currently lacks a public API to track ad events on Android. This change enables KMP applications to track ad impressions, revenue, and other ad-related events through the RevenueCat SDK.

### Description

This PR exposes the ad tracking APIs from the Android SDK to KMP applications. The implementation includes:

**API Changes:**
- Added `@ExperimentalRevenueCatApi` annotation to mark these APIs as experimental
- Added 7 new data types in the `models` module:
  - `AdMediatorName` - Type-safe wrapper for ad mediator names (AdMob, AppLovin, etc.)
  - `AdRevenuePrecision` - Type-safe wrapper for revenue precision (exact, estimated, etc.)
  - `AdDisplayedData` - Data for tracking ad impressions
  - `AdOpenedData` - Data for tracking ad clicks
  - `AdRevenueData` - Data for tracking ad revenue
  - `AdLoadedData` - Data for tracking successful ad loads
  - `AdFailedToLoadData` - Data for tracking failed ad loads
- Added 5 tracking methods to the `Purchases` class:
  - `trackAdDisplayed(data: AdDisplayedData)`
  - `trackAdOpened(data: AdOpenedData)`
  - `trackAdRevenue(data: AdRevenueData)`
  - `trackAdLoaded(data: AdLoadedData)`
  - `trackAdFailedToLoad(data: AdFailedToLoadData)`

**Implementation Details:**
- Android implementation delegates to `androidPurchases.adTracker`
- iOS implementation throws error with message "Ad tracking is not yet supported in iOS KMP"
- All APIs are marked with `@ExperimentalRevenueCatApi` to indicate experimental status
- Value classes use public constructors for flexibility while maintaining type safety
- Updated Android API baseline to include new public APIs

**Testing:**
- Added unit tests for all ad event data types in `models/src/commonTest`
- Added mapping tests for Android type conversions in `mappings/src/androidInstrumentedTest`
- All tests verify construction, property access, and type conversions
- Android compilation, API checks, unit tests, and lint checks all passing

**Commits:**
1. Add ExperimentalRevenueCatApi annotation
2. Add ad event data types to KMP models  
3. Expose ad tracking methods in KMP Purchases class
4. Add unit tests for ad tracking types and mappings
5. Update API baseline for ad tracking types


models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdRevenuePrecision.kt
```diff
@@ -0,0 +1,18 @@
...
+@ExperimentalRevenueCatApi
+@JvmInline
+public value class AdRevenuePrecision(public val value: String) {
+    public companion object {
+        public val EXACT: AdRevenuePrecision = AdRevenuePrecision("exact")
+        public val PUBLISHER_DEFINED: AdRevenuePrecision = AdRevenuePrecision("publisher_defined")
+        public val ESTIMATED: AdRevenuePrecision = AdRevenuePrecision("estimated")
+        public val UNKNOWN: AdRevenuePrecision = AdRevenuePrecision("unknown")
+    }
+}
```
skydoves, line 45: Would be good to have more detailed docs for each precision.

skydoves (approved):
Overall looks good to me, just minor comments for the docs.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
```diff
@@ -815,6 +820,46 @@ public expect class Purchases {
...
      */
     public fun getCachedVirtualCurrencies(): VirtualCurrencies?
+
+    /**
+     * Tracks an ad displayed event.
+     *
+     * @param data The ad display event data.
+     */
+    @ExperimentalRevenueCatApi
+    public fun trackAdDisplayed(data: AdDisplayedData)
```
tonidero: Should we follow the same structure like in the native SDKs, and create an `AdTracker` class that will have these methods? I understand that pattern is not something we've done before in the KMP SDK, but I think it could be good to keep a cleaner API

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -790,4 +795,24 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
         val cachedVirtualCurrencies: RCVirtualCurrencies? = iosPurchases.cachedVirtualCurrencies()
         return cachedVirtualCurrencies?.toVirtualCurrencies()
     }
+
+    @ExperimentalRevenueCatApi
+    public actual fun trackAdDisplayed(data: AdDisplayedData): Unit =
+        error("Ad tracking is not yet supported in iOS KMP")
```
tonidero: Just curious, is there any reason we're not adding support for the iOS SDKs? In general, we try to avoid throwing exceptions like this as much as possible since it could potentially cause crashes that we would like to avoid.

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -577,6 +583,31 @@ public actual class Purchases private constructor(private val androidPurchases:
         return cachedVirtualCurrencies?.toVirtualCurrencies()
     }
 
+    @OptIn(ExperimentalRevenueCatApi::class, com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI::class)
```
tonidero: I see that in the iOS implementation, we propagate the `@ExperimentalRevenueCatApi` instead of opt in. I think that makes more sense here too?

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
polmiro: I'm fast-tracking Android for a customer using KMP only with Android that will be trying out the feature. But the plan is to follow up soon with the iOS side. Since the feature is still experimental and invite-only I think it's okay to release first the Android one?

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
polmiro: I actually considered this but seeing everything was flat and mostly delegating to Android I decided not to. I'll go for it now.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
tonidero: Note that people may still call these APIs even if invite-only. But it's true that this is still an experimental API, so I won't block this PR if we want to get this ASAP. But I think we should address soon after.

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
tonidero: Yeah agreed that's how we've ben doing things, but feels better to keep the same API if we can, and avoid polluting the `Purchases` interface too much.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdRevenuePrecision.kt
polmiro: Good call, addressed!

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
polmiro: Sounds good, I have a ticket to track this and will make sure we tackle it next.

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
polmiro: Fixed!

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
polmiro: Fixed, if you can take another look to make sure it looks like what you expected. Thanks

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/AdTracker.android.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.revenuecat.purchases.kmp.models.AdOpenedData
+import com.revenuecat.purchases.kmp.models.AdRevenueData
+
+/**
+ * Android implementation of [AdTracker] that delegates to the native Android SDK.
+ */
+@ExperimentalRevenueCatApi
+@OptIn(com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI::class)
+public actual class AdTracker internal constructor(
+    private val androidAdTracker: com.revenuecat.purchases.ads.events.AdTracker
```
tonidero: Probably nicer to add a named import like `AndroidAdTracker` instead of the fully qualified name here.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ExperimentalRevenueCatApi.kt
```diff
@@ -0,0 +1,20 @@
...
+    level = RequiresOptIn.Level.WARNING
+)
+@Retention(AnnotationRetention.BINARY)
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY,
+    AnnotationTarget.TYPEALIAS
+)
+public annotation class ExperimentalRevenueCatApi
```
tonidero, line 20: I'm wondering if we should reuse the same name as the annotation we have in Android... But I guess that, even if we have the same name, the package name would be different and this way it's a bit easier to differentiate them... So I guess I'm ok keeping them different.

tonidero (approved):
Looks great to me!

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/AdTracker.android.kt
polmiro: Fixed! also found two more occurences and addressed them.

PR #642, opened 2026-01-21
Convert ad event model data classes to regular classes

polmiro:
## Summary
Convert `AdDisplayedData`, `AdFailedToLoadData`, `AdLoadedData`, `AdOpenedData`, and `AdRevenueData` from data classes to regular classes to avoid binary compatibility issues in the public API.

This addresses feedback from the initial ad events PR about avoiding data classes in public APIs.

## Context
Data classes automatically generate `copy()` and `componentN()` functions that cause binary incompatibility when adding new properties. Following Kotlin API guidelines and best practices, these model classes are now regular classes without the auto-generated methods.

The classes retain all properties and constructors. Methods like `equals()`, `hashCode()`, and `toString()` can be added manually later if needed.

## Changes
- Changed `data class` to `class` for all 5 ad event model types
- Added explicit `public` visibility modifiers to properties (required by explicit API mode)
- Updated API dumps to reflect removal of:
  - `copy()` functions
  - `componentN()` destructuring functions
  - Auto-generated `equals()`, `hashCode()`, `toString()` methods

## Test plan
- [x] All existing tests pass
- [x] API dumps updated and verified
- [x] Models module builds successfully

## References
- [Kotlin API Guidelines: Avoid using data classes in your API](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api)
- [Public API Challenges in Kotlin by Jake Wharton](https://jakewharton.com/public-api-challenges-in-kotlin/)


tonidero (approved):
Ohhhh right I totally missed that 🤦 Thanks for fixing it! I was thinking whether we should add equals/toString to these (maybe by using `Poko` like in Android) but I think it's not needed right now.

JayShortway (approved):
Thanks for doing this! 🙏

PR #648, opened 2026-01-23
Add iOS ad tracking API support

polmiro:
## Summary
This PR adds iOS ad tracking API support to the KMP SDK, bridging to the native iOS implementation through PurchasesHybridCommon.

### Core Implementation
- Added `AdTracker` interface and iOS implementation
- Added ad tracking data models with `adFormat` parameter:
  - `AdDisplayedData`
  - `AdOpenedData`
  - `AdRevenueData`
  - `AdLoadedData`
  - `AdFailedToLoadData`
- Added `AdFormat` value class with common format types (BANNER, INTERSTITIAL, REWARDED, etc.)
- Added `ExperimentalRevenueCatApi` annotation for marking experimental APIs

### API Availability
- Ad tracking APIs require iOS 15.0+ / tvOS 15.0+ / macOS 12.0+ / watchOS 8.0+

### Testing
- Added ad tracking testing screen in composeApp sample

## Dependencies
- Requires PurchasesHybridCommon 17.30.1

## Notes
- All ad tracking APIs are marked with `@ExperimentalRevenueCatApi`
- `adFormat` is passed to iOS but currently ignored by PHC iOS (pending iOS SDK support)

### DIFF
--- a/core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/AdTracker.ios.kt
+++ b/core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/AdTracker.ios.kt
@@ -15,6 +15,17 @@ import com.revenuecat.purchases.kmp.models.AdRevenueData
 import kotlinx.cinterop.ExperimentalForeignApi
 import platform.Foundation.NSNumber
 
+private const val KEY_NETWORK_NAME = "networkName"
+private const val KEY_MEDIATOR_NAME = "mediatorName"
+private const val KEY_AD_FORMAT = "adFormat"
+private const val KEY_PLACEMENT = "placement"
+private const val KEY_AD_UNIT_ID = "adUnitId"
+private const val KEY_IMPRESSION_ID = "impressionId"
+private const val KEY_REVENUE_MICROS = "revenueMicros"
+private const val KEY_CURRENCY = "currency"
+private const val KEY_PRECISION = "precision"
+private const val KEY_MEDIATOR_ERROR_CODE = "mediatorErrorCode"
+
 @ExperimentalRevenueCatApi
 @OptIn(ExperimentalForeignApi::class)
 public actual class AdTracker {
@@ -25,12 +36,12 @@ public actual class AdTracker {
         }
 
         val adData = mapOf<Any?, Any?>(
-            "networkName" to data.networkName,
-            "mediatorName" to data.mediatorName.value,
-            "adFormat" to data.adFormat.value,
-            "placement" to data.placement,
-            "adUnitId" to data.adUnitId,
-            "impressionId" to data.impressionId
+            KEY_NETWORK_NAME to data.networkName,
+            KEY_MEDIATOR_NAME to data.mediatorName.value,
+            KEY_AD_FORMAT to data.adFormat.value,
+            KEY_PLACEMENT to data.placement,
+            KEY_AD_UNIT_ID to data.adUnitId,
+            KEY_IMPRESSION_ID to data.impressionId
         )
 
         RCCommonFunctionality.trackAdDisplayed(adData)
@@ -43,12 +54,12 @@ public actual class AdTracker {
         }
 
         val adData = mapOf<Any?, Any?>(
-            "networkName" to data.networkName,
-            "mediatorName" to data.mediatorName.value,
-            "adFormat" to data.adFormat.value,
-            "placement" to data.placement,
-            "adUnitId" to data.adUnitId,
-            "impressionId" to data.impressionId
+            KEY_NETWORK_NAME to data.networkName,
+            KEY_MEDIATOR_NAME to data.mediatorName.value,
+            KEY_AD_FORMAT to data.adFormat.value,
+            KEY_PLACEMENT to data.placement,
+            KEY_AD_UNIT_ID to data.adUnitId,
+            KEY_IMPRESSION_ID to data.impressionId
         )
 
         RCCommonFunctionality.trackAdOpened(adData)
@@ -61,15 +72,15 @@ public actual class AdTracker {
         }
 
         val adData = mapOf<Any?, Any?>(
-            "networkName" to data.networkName,
-            "mediatorName" to data.mediatorName.value,
-            "adFormat" to data.adFormat.value,
-            "placement" to data.placement,
-            "adUnitId" to data.adUnitId,
-            "impressionId" to data.impressionId,
-            "revenueMicros" to NSNumber(long = data.revenueMicros),
-            "currency" to data.currency,
-            "precision" to data.precision.value
+            KEY_NETWORK_NAME to data.networkName,
+            KEY_MEDIATOR_NAME to data.mediatorName.value,
+            KEY_AD_FORMAT to data.adFormat.value,
+            KEY_PLACEMENT to data.placement,
+            KEY_AD_UNIT_ID to data.adUnitId,
+            KEY_IMPRESSION_ID to data.impressionId,
+            KEY_REVENUE_MICROS to NSNumber(long = data.revenueMicros),
+            KEY_CURRENCY to data.currency,
+            KEY_PRECISION to data.precision.value
         )
 
         RCCommonFunctionality.trackAdRevenue(adData)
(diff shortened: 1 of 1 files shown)


core/build.gradle.kts
```diff
@@ -45,6 +45,10 @@ kotlin {
         pod("PurchasesHybridCommon") {
             version = libs.versions.revenuecat.common.get()
             extraOpts += listOf("-compiler-option", "-fmodules")
+            // FIXME Undo this before merging:
+            source = git("https://github.com/RevenueCat/purchases-hybrid-common") {
+                branch = "add-ios-ad-tracking-bridge"
+            }
```
JayShortway: Attaching a comment here so we don't forget to revert this.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import com.revenuecat.purchases.kmp.ExperimentalRevenueCatApi
+import kotlin.jvm.JvmInline
+
+/**
+ * Common ad format types.
+ */
+@ExperimentalRevenueCatApi
+@JvmInline
+public value class AdFormat(public val value: String) {
```
JayShortway: We should be careful with value classes in the public API, as it causes name-mangling on the Java side.

For instance, if you check [models.api](https://github.com/RevenueCat/purchases-kmp/pull/648/changes#diff-44c3c1848c7380b7140dfdc8bd00a7b361e940989bf9b0f65cf3403854f406d2R6), it shows:

```java
public final class com/revenuecat/purchases/kmp/models/AdDisplayedData {
	// ...
	public final fun getAdFormat-AVhMwwk ()Ljava/lang/String;
```

The `-AVhMwwk` part is added because `AdFormat` is a value class.

We could try to use `@JvmName("getAdFormat")` everywhere we declare the `adFormat: AdFormat` property, but it's easy to forget (and I'm not sure if it will work, but worth a shot).

Alternatively we can avoid public value classes altogether and just use regular classes.

Some more details in [the docs](https://kotlinlang.org/docs/inline-classes.html#mangling).

polmiro: Using regular classes seems like the simplest. Going for that! Thank you.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdMediatorName.kt
```diff
@@ -1,14 +1,12 @@
...
 import com.revenuecat.purchases.kmp.ExperimentalRevenueCatApi
-import kotlin.jvm.JvmInline
 
 /**
  * Common ad mediator names.
  */
 @ExperimentalRevenueCatApi
-@JvmInline
-public value class AdMediatorName(public val value: String) {
+public class AdMediatorName(public val value: String) {
```
polmiro, line 9: Spotted this one too @JayShortway . I fixed it since it's experimental.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdRevenuePrecision.kt
```diff
@@ -10,8 +9,7 @@ import kotlin.jvm.JvmInline
  * This enum helps distinguish between exact reported values and estimates.
  */
 @ExperimentalRevenueCatApi
-@JvmInline
-public value class AdRevenuePrecision(public val value: String) {
+public class AdRevenuePrecision(public val value: String) {
```
polmiro, line 12: And this one.

iosApp/Podfile
```diff
@@ -6,6 +6,6 @@ target 'iosApp' do
   # use_frameworks!
 
   # Pods for iosApp
-  pod 'PurchasesHybridCommon', '17.30.0'
-  pod 'PurchasesHybridCommonUI', '17.30.0'
+  pod 'PurchasesHybridCommon', '17.30.1'
+  pod 'PurchasesHybridCommonUI', '17.30.1'
```
JayShortway: Preferably the PHC update is done in a separate PR for changelog reasons.

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/AdTrackingTestingScreen.kt
```diff
@@ -0,0 +1,321 @@
...
+            )
+            Purchases.sharedInstance.adTracker.trackAdDisplayed(data)
+            statusMessage = "Ad displayed event tracked successfully!"
+            lastTrackedFunction = AdTrackingFunction.TRACK_AD_DISPLAYED
+            messageColor = Color.Green
+        } catch (e: Exception) {
+            statusMessage = "Error: ${e.message}"
+            lastTrackedFunction = AdTrackingFunction.TRACK_AD_DISPLAYED
+            messageColor = Color.Red
+        }
```
JayShortway: Can these tracking functions throw exceptions? If so, we might want to try to avoid that. If not, we can remove the `try catch` imo.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/AdTracker.ios.kt
```diff
@@ -1,28 +1,116 @@
...
 import com.revenuecat.purchases.kmp.models.AdFailedToLoadData
 import com.revenuecat.purchases.kmp.models.AdLoadedData
 import com.revenuecat.purchases.kmp.models.AdOpenedData
 import com.revenuecat.purchases.kmp.models.AdRevenueData
+import kotlinx.cinterop.ExperimentalForeignApi
+import platform.Foundation.NSNumber
 
 /**
- * iOS implementation of [AdTracker]. Ad tracking is not yet supported on iOS.
+ * iOS implementation of [AdTracker] that bridges to CommonFunctionality.
```
JayShortway: This is public documentation, so it shouldn't mention implementation details. 

For now it would be fine to omit the docs here entirely, as we have only documented the `commonMain` source set, but it is definitely nicer to document the platform source sets too (`iosMain` and `androidMain`).

iosApp/Podfile
tonidero: I think it needs to be in the same PR to keep tests passing since there was a breaking change :( But we have #662 so we might want to bump to the latest.

polmiro: I upgraded to latest `17.31.0`. Sorry it got a bit messy with the ad format breaking change.

revenuecatui/build.gradle.kts
```diff
@@ -32,6 +32,10 @@ kotlin {
         version = libs.versions.revenuecat.kmp.get()
         ios.deploymentTarget = libs.versions.ios.deploymentTarget.ui.get()
 
+        pod("PurchasesHybridCommon") {
```
tonidero: We were chatting about this over Slack, but I'm very confused as to why this is needed after these changes... Any ideas @JayShortway ? It seems that it's not finding the purchases-ios native library symbols anymore using the previous naming, and we need to add this and find it through cocoapods... but not sure why it was able to find those symbols before and not not 😕

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/AdTracker.ios.kt
```diff
@@ -1,28 +1,113 @@
...
+    }
+
+    public actual fun trackAdRevenue(data: AdRevenueData) {
+        if (!IOSAPIAvailabilityChecker().isAdTrackingAPIAvailable()) {
+            Purchases.logHandler.w("Purchases", "Ad tracking requires iOS 15.0+. Current API is unavailable.")
+            return
+        }
+
+        val adData = mapOf<Any?, Any?>(
+            "networkName" to data.networkName,
```
tonidero: Since many of these strings are actually repeated, could be worth extracting to constants

revenuecatui/build.gradle.kts
tonidero: We talked over Slack, we can go with the current approach for now, but this is probably not needed. Can you confirm @polmiro ?

JayShortway: Yes, let's remove this before merging.

polmiro: Just updated! Thanks for double-checking on this.

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/AdTracker.ios.kt
polmiro: Done!

PR #660, opened 2026-02-02
Adds support for Swift resources

JayShortway:
## Motivation
I'm working on improving the integration experience of purchases-kmp, by avoiding the need for app developers to add the purchases-hybrid-common dependency manually as this adds friction and prevents properly testing Kotlin/Native modules that depend on purchases-kmp. Instead, we will be embedding purchases-ios directly in purchases-kmp, handling all complexity for the app developers. This PR adds support for bundling Swift resources, such as strings and icons.

## How it works
1. When a `swiftPackage` dependency is added, any resources it declares are automatically discovered from the package info. 
2. Then, the new `ProcessSwiftResourcesTask` will run, which:
    a. compiles `.xcassets` into an asset catalog with `actool`,
    b. places localized strings in the correct folder structure, and
    c. copies over any other files.
3. These processed resources are then picked up by Compose Resources, which will include them in the final published `.klib`.

### DIFF
--- a/build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/SwiftResourcesOutputTests.kt
+++ b/build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/SwiftResourcesOutputTests.kt
@@ -1,0 +1,92 @@
+package com.revenuecat.purchases.kmp.buildlogic
+
+import com.revenuecat.purchases.kmp.buildlogic.test.SwiftPackageBuilder
+import com.revenuecat.purchases.kmp.buildlogic.test.revenueCatLibraryPluginTest
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.condition.EnabledOnOs
+import org.junit.jupiter.api.condition.OS
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+
+/**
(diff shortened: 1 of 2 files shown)


build-logic/convention/src/test/kotlin/com/revenuecat/purchases/kmp/buildlogic/SwiftResourcesUpToDateTests.kt
```diff
@@ -0,0 +1,122 @@
...
+
+
+/**
+ * Tests that verify ProcessSwiftResourcesTask reruns when resource files change.
+ */
+@EnabledOnOs(OS.MAC)
+class SwiftResourcesUpToDateTests {
+
+    @Test
+    fun `processSwiftResources re-runs when resource file is added`() =
```
rickvdl, line 19: Can we / should we also add some tests asserting on the output of the newly added task? E.g. asserting that `.copy` resources are just copied, localized strings are put in the right place and the asset catalog is compiled?

JayShortway: That's a great idea! Added those tests here: https://github.com/RevenueCat/purchases-kmp/pull/660/commits/b949498b03998720edf34466d746d050ebe0bb59.

PR #663, opened 2026-02-04
Adds support for Swift compiler flags

JayShortway:
## Motivation
I'm working on improving the integration experience of purchases-kmp, by avoiding the need for app developers to add the purchases-hybrid-common dependency manually as this adds friction and prevents properly testing Kotlin/Native modules that depend on purchases-kmp. Instead, we will be embedding purchases-ios directly in purchases-kmp, handling all complexity for the app developers. This PR adds support for Swift compiler flags. We will eventually need this to allow purchases-ios to find the (Compose) resources.

## How it works
1. A `swiftPackage` dependency can declare optional `SwiftSettings`. Only `define` is supported for now.
```kotlin
swiftPackage(
    // ...
    swiftSettings = SwiftSettings {
        define("COMPOSE_RESOURCES")
    }
)
```
2. These get converted to Swift compiler arguments (`-Xswiftc`), and
3. passed to the `SwiftBuildTask`.

### DIFF
--- a/composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/AdTrackingTestingScreen.kt
+++ b/composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/AdTrackingTestingScreen.kt
@@ -1,0 +1,291 @@
+package com.revenuecat.purchases.kmp.sample
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.Button
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.Text
+import androidx.compose.material.TextButton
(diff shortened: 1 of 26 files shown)


build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/model/SwiftSettings.kt
```diff
@@ -0,0 +1,47 @@
...
+    fun define(name: String) {
+        defines.add(name)
+    }
+    
+    /**
+     * Converts the settings to command line arguments for `swift build`.
+     * Each `define` becomes `-Xswiftc -D -Xswiftc FLAG_NAME`.
+     */
+    internal fun toCommandLineArgs(): List<String> =
+        defines.flatMap { listOf("-Xswiftc", "-D", "-Xswiftc", it) }
```
rickvdl, line 37: IIRC this can (or should?) be one argument, so `-Xswiftc -D{flag}`. Interesting if this works as well

rickvdl (commented):
Nice addition!

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/model/SwiftSettings.kt
JayShortway: My understanding is that both work, because `swiftc` can receive flags in 2 forms: 
- `-D FLAG`
- `-DFLAG`

The former is what I'm doing currently, which I've tested and seems to work fine.

rickvdl: Should be fine indeed then :)

PR #668, opened 2026-02-06
Adjusts the package name of generated Kotlin bindings

JayShortway:
## Motivation

## Changes
This PR only changes the package name of the generated Kotlin bindings in preparation for SPM import. This helps keep the follow-up PRs smaller. No logic has changed in this PR. These are internal changes only.

### DIFF
--- a/mappings/build.gradle.kts
+++ b/mappings/build.gradle.kts
@@ -27,6 +27,7 @@ kotlin {
         pod("PurchasesHybridCommon") {
             version = libs.versions.revenuecat.common.get()
             extraOpts += listOf("-compiler-option", "-fmodules")
+            packageName = "swiftPMImport.com.revenuecat.purchases.kn.core"
         }
     }
 }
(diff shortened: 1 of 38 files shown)


mappings/build.gradle.kts
```diff
@@ -27,6 +27,7 @@ kotlin {
         pod("PurchasesHybridCommon") {
             version = libs.versions.revenuecat.common.get()
             extraOpts += listOf("-compiler-option", "-fmodules")
+            packageName = "swiftPMImport.com.revenuecat.purchases.kn.core"
```
tonidero, line 30: Just wondering, should these change the name for each module? Like this being 
```suggestion
            packageName = "swiftPMImport.com.revenuecat.purchases.kn.mappings"
```

We weren't differentiating for cocoapods, so it might be ok... Just wondering if it was worth changing it now. Same for the models module (though no changes appeared happened there)

tonidero (approved):
Just a question, but I think it makes sense. This will definitely help with the transition, thanks for making this easier to review 🙏

mappings/build.gradle.kts
JayShortway: Ah no, because this package name is where the Kotlin bindings for `PurchasesHybridCommon` will end up. So if we would differentiate per module, we'd have the _same_ (duplicate) Kotlin bindings in different packages (e.g. in `*.core`, `*.mappings` and `*.models`.

PR #698, opened 2026-02-23
feat: add convenience method for canMakePayments

### LINKED ISSUE #689: Provide suspend equivalent for `canMakePayments`
Very small feature request, but it would be useful to have a `suspend` equivalent of the `canMakePayments` function that's present on the companion object of `Purchases` right now, that currently takes a callback.

I also realise that the issue template for feature requests suggests using your website form, but this felt small and relevant enough to the repo itself to report it here - hopefully this is okay!

Thanks!

TheRogue76:
### Checklist

- [x] If applicable, create follow-up issues for `purchases-android`, `purchases-ios` and other
  hybrids

### Motivation

In most cases, people prefer using suspending functions to callbacks (mostly for chaining calls), so i figured it would be good to add the method for it.

admittedly i do feel like there being two methods doing the same thing is a little bit of a smell especially for something this small, but removing the non suspending call is a breaking change, and the android SDK very much uses callbacks, and i would like to get some feedback if this is a route that would make sense to pursue or not, and if so, i can do a PR on Android too.
Resolves #689

### Description

Add a convenience suspending function for the `canMakePayments` method on the SDK called `awaitCanMakePayments`
Re-ran apiDump to make sure the list of APIs is up to date
No specific validations done, as there is not much to the API

### DIFF
--- a/core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
+++ b/core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
@@ -69,8 +69,6 @@ import com.revenuecat.purchases.restorePurchasesWith
 import com.revenuecat.purchases.syncAttributesAndOfferingsIfNeededWith
 import com.revenuecat.purchases.syncPurchasesWith
 import java.net.URL
-import kotlin.coroutines.resume
-import kotlin.coroutines.suspendCoroutine
 import com.revenuecat.purchases.DangerousSettings as AndroidDangerousSettings
 import com.revenuecat.purchases.Purchases as AndroidPurchases
 import com.revenuecat.purchases.hybridcommon.configure as commonConfigure
--- a/core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
+++ b/core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
@@ -67,8 +67,6 @@ import com.revenuecat.purchases.kmp.models.WinBackOffer
 import com.revenuecat.purchases.kmp.strings.ConfigureStrings
 import platform.Foundation.NSError
 import platform.Foundation.NSURL
-import kotlin.coroutines.resume
-import kotlin.coroutines.suspendCoroutine
 import swiftPMImport.com.revenuecat.purchases.kn.core.RCDangerousSettings as IosDangerousSettings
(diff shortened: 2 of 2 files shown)


core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
```diff
@@ -104,6 +104,25 @@ public expect class Purchases {
...
+         * are supported. This method is asynchronous since it requires a connected BillingClient.
+         *
+         * This is a convenience method for [canMakePayments] that wraps the callback via a suspending coroutine
+         *
+         * @param features Play Store only. A list of feature types to check for support. Feature
+         * types must be one of [BillingFeature]. By default, is an empty list and no specific
+         * feature support will be checked.
+         * @return result of the check
+         */
+        public suspend fun canMakePayments(
```
tonidero: Following our normal syntax for coroutines versions of APIs, I think we should:
- Move this function to be an extension function on `Coroutines.kt`
- Rename this function to `awaitCanMakePayments`

Lmk what you think :)

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -148,6 +150,13 @@ public actual class Purchases private constructor(private val androidPurchases:
             features = features.map { it.toAndroidBillingFeature() },
         ) { result -> callback(result) }
 
+        @JvmStatic
+        public actual suspend fun canMakePayments(
+            features: List<BillingFeature>
+        ): Boolean = suspendCoroutine { continuation ->
+            canMakePayments(features = features, callback = continuation::resume)
```
tonidero: Indeed, could be good to add a coroutine version of this in purchases-android and have KMP use that directly... But I think that can also come later/as a follow-up.

TheRogue76: Sounds good. Would be nice

core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
TheRogue76: Oh this is dope, i didn't know that existed. Updated the implementation under [fb19543](https://github.com/RevenueCat/purchases-kmp/pull/698/commits/fb195438b24d3f1c4862d78f410cddabec5de0be)

tonidero (approved):
Looking great thanks! We will get this merged and will release it soon :)

PR #701, opened 2026-02-24
Add CircleCI job and Fastlane lane for updating API files

tonidero:
## Summary
- Adds an `update_api_files` Fastlane lane that runs `apiDump`, detects changed API files (`*.api` / `*.klib.api`), and creates a PR with the updates against the triggering branch
- Adds a corresponding `update-api-files` CircleCI job and `on-action-update-api-files` workflow, triggered via the `action: update-api-files` pipeline parameter
- Removes the need for developers to manually run `apiDump` and commit results when `apiCheck` fails

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tonidero:
In the end, I think this is not needed... Faster to do it locally in the end (I originally created it due to some incompatibility issues between my Xcode version and KMP version, but I can work around that.

PR #708, opened 2026-02-27
Add maestro E2E test app

ajpallares:
## Summary

Adds a Maestro E2E test app under `e2e-tests/MaestroTestApp/` — a minimal Kotlin Multiplatform Compose app (Android + iOS) used by automated Maestro flows to verify the RevenueCat purchase integration.

- Two screens: "Test Cases" list and "Purchase through paywall" (presents a RevenueCat V2 paywall and displays entitlement status)
- App ID `com.revenuecat.automatedsdktests`
- API key placeholder (`MAESTRO_TESTS_REVENUECAT_API_KEY`) replaced at CI time via `sed`
- Local SDK resolution via Gradle project references (`:purchases` and `:ui:revenuecatui`)
- Errors from `getCustomerInfo` and paywall presentation are surfaced in the UI (not just logged) for Maestro screenshot capture
- Compose `testTag` modifiers for reliable Maestro element targeting
- **iOS paywall presented as a native modal** instead of inline Compose `Paywall(options)`. Compose Multiplatform's `UIKitViewController` accessibility bridge (`isNativeAccessibilityEnabled = true`) does not propagate the embedded SwiftUI paywall's accessibility tree to the system, making it invisible to Maestro. Presenting as a native modal bypasses this limitation. Android uses the standard inline Compose paywall, which works correctly with Maestro.
- Includes a README

Counterpart PRs: [react-native-purchases#1635](https://github.com/RevenueCat/react-native-purchases/pull/1635), [purchases-flutter#1654](https://github.com/RevenueCat/purchases-flutter/pull/1654), [purchases-capacitor#699](https://github.com/RevenueCat/purchases-capacitor/pull/699), [cordova-plugin-purchases#857](https://github.com/RevenueCat/cordova-plugin-purchases/pull/857), [purchases-unity#836](https://github.com/RevenueCat/purchases-unity/pull/836)

### Follow-up PRs (stacked)
- [#709](https://github.com/RevenueCat/purchases-kmp/pull/709) — Maestro test flows (YAML)
- [#710](https://github.com/RevenueCat/purchases-kmp/pull/710) — CircleCI jobs to run the tests


e2e-tests/MaestroTestApp/README.md
JayShortway, line 1: (Not really related to this file, I just want a thread 😄) 

I was working on a Maestro setup as well here: https://github.com/RevenueCat/purchases-kmp/pull/704. In my case, the purpose is to check local resources are included correctly by rendering 2 paywalls: 
- the fallback paywall
- a legacy paywall with icons

Maybe we can think about how we can best combine our plans? To be clear, we do not necessarily need to use the same test app (only if we think it's helpful), but maybe we can align on:
- parent folder: in my PR it's `test-apps`, in this PR it's `e2e-tests`. I chose `test-apps` because I can see us using apps for other kind of tests, such as binary size or benchmarks. 
- integrate the SDK via Maven Local: I did that in my PR because that influences resource loading. Might be good to do that always, just in case? The only thing is that I still wanted to optimize the pipeline further, as all combined it adds quite a lot of runtime. 
- maybe some other things?

ajpallares: I agree! We should totally try to unify.

I agree with using the same parent folder.

Regarding using the same app or not. I think your use case can totally fit the MaestroTestApp's purpose. Your 2 tests can be 2 more tests in the MaestroTestApp list. It would probably be good to unify the app structure (navigation and screen) with the MaestroTestApp in other hybrids. And in fact, the 2 test cases you're working on in #704 may be useful to replicate in the other hybrids as well.

ajpallares:
Superseded — rebased onto `3.0.0-dev` as the base branch in #824.

PR #714, opened 2026-03-03
Adds facade Gradle modules for Swift packages

JayShortway:
## Motivation

## Changes
This PR adds 2 facade modules: 
- `kn-core`
- `kn-ui`
  
These are Kotlin/Native facades for the `RevenueCat` and `RevenueCatUI` Swift packages respectively. These Swift packages are added as dependencies using our own `swiftPackage()`. This means that these modules will contain the automatically generated Kotlin bindings for this Swift code. The modules aren't used yet.

## Note
This PR targets a new `3.0.0-dev` branch. The facade modules aren't used, but they would be published if the release train runs. The integration changes will require a major, which I would like to release as alpha/beta first. So that's why I created a separate parallel branch. I will ensure that `3.0.0-dev` stays up to date with `main`.


build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/ConfigureSwiftDependencies.kt
```diff
@@ -26,6 +27,10 @@ fun Project.configureSwiftDependencies() {
 
     afterEvaluate {
         if (registry.packages.isNotEmpty()) {
+            if (!OperatingSystem.current().isMacOsX) {
+                logger.info("Skipping Swift dependency configuration on non-macOS host")
+                return@afterEvaluate
+            }
```
JayShortway, line 33: Added this to allow some of our CI jobs to still be run on Docker executors.

.circleci/config.yml
```diff
@@ -158,9 +171,10 @@ jobs:
           command: bundle exec fastlane prepare_next_snapshot_version
 
   detekt:
-    executor: jdk17
+    executor: xcode16
```
JayShortway, line 174: This is probably not needed, as afaik detekt runs on JVM, but I figured better safe than sorry.

tonidero: I guess the only thing is that it might delay this job when we don't have resources in CircleCI... but I guess it should be normally ok 🤞

kn-core/src/swift/Package.swift
```diff
@@ -0,0 +1,20 @@
+// swift-tools-version:5.9
+import PackageDescription
+
+let package = Package(
+    name: "AdditionalSwift",
+    platforms: [
+        .macOS(.v10_15),
+        .watchOS("6.2"),
```
tonidero, line 8: Huh, interesting that watchos is the only that needs to be a string?

kn-core/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+        iosMain.dependencies {
+            swiftPackage(
+                path = rootProject.file("upstream/purchases-ios"),
+                target = "RevenueCat",
+                packageName = "swiftPMImport.com.revenuecat.purchases.kn.core",
+                customDeclarations = """
+                    // Force cinterop binding generation for types otherwise not in the public API
+                    static inline int __forceBindings(
+                        enum RCStoreMessageType _1
+                    ) { return 0; }
```
tonidero, line 18: I had to think about this for a while 😅

kn-core/src/swift/Package.swift
JayShortway: Yea, it's the same in [purchases-ios](https://github.com/RevenueCat/purchases-ios/blob/3baeae273b747935d2184e903029b9f6e1de1223/Package.swift#L73). Apparently there's no enum case for `6.2`.

.circleci/config.yml
JayShortway: Yea indeed. We can keep an eye out.

PR #715, opened 2026-03-04
Add PurchaseLogic support to PaywallView

rickvdl:
## Summary
- Adds `PaywallPurchaseLogic` interface and `PurchaseLogicResult` sealed class, allowing developers to provide custom purchase and restore logic to `PaywallView`
- Adds `purchaseLogic` property to `PaywallOptions`
- Implements iOS support via `PaywallProxy` + `HybridPurchaseLogicBridge`
- Implements Android support via `performPurchase`/`performRestore` lambdas
- Adds `SamplePurchaseLogic` to the sample app with a compile-time `purchasesAreCompletedByMyApp` flag
- Adds API tester coverage for the new public API surface

### Related PRs
- Flutter: RevenueCat/purchases-flutter#1657
- React Native: RevenueCat/react-native-purchases#1607

## Test plan
- [x] Build and run on iOS simulator — verify paywall displays and purchase logic button appears
- [x] Enable `purchasesAreCompletedByMyApp` in `Defaults.kt` and verify custom purchase logic is invoked
- [x] Build and run on Android — verify paywall displays with purchase logic support
- [x] Verify standard paywall flow (without purchase logic) is unaffected


composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/SamplePurchaseLogic.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import com.revenuecat.purchases.kmp.models.Package
+import com.revenuecat.purchases.kmp.ui.revenuecatui.PaywallPurchaseLogic
+import com.revenuecat.purchases.kmp.ui.revenuecatui.PurchaseLogicResult
+import kotlinx.coroutines.delay
+
+internal class SamplePurchaseLogic : PaywallPurchaseLogic {
+    override suspend fun performPurchase(rcPackage: Package): PurchaseLogicResult {
+        println("[SamplePurchaseLogic] performPurchase called for: ${rcPackage.identifier}")
+        delay(2000)
```
tonidero: I do wonder if we should just call our own purchase/restore methods in these for a more realistic test... But not a blocker.

revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Performs an in-app purchase for the specified package.
+     *
+     * If a purchase is successful, `syncPurchases` will automatically be called by RevenueCat.
+     *
+     * @param rcPackage The package representing the in-app product that the user intends
+     * to purchase.
+     * @return A [PurchaseLogicResult] indicating the outcome of the purchase operation.
+     */
+    public suspend fun performPurchase(rcPackage: Package): PurchaseLogicResult
```
tonidero: So I remember now that in Android, we changed this to receive a `PaywallPurchaseLogicParams`, that will enable other options like setting a product change information (for Android only). We probably want to do the same here? (We might want to do the same in all hybrids, but I missed that :(, and I should do the same in the Unity PR I have open 🥴)

revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Restores previously completed purchases.
+     *
+     * If restoration is successful, `syncPurchases` will automatically be called by RevenueCat.
+     *
+     * @return A [PurchaseLogicResult] indicating the outcome of the restoration process.
+     */
+    public suspend fun performRestore(): PurchaseLogicResult
```
tonidero: I guess in android we pass the CustomerInfo as parameter here... I don't fully remember the logic for that though...

revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Represents the result of a purchase or restore attempt made by custom app-based code
+ * (not RevenueCat).
+ */
+public sealed interface PurchaseLogicResult {
```
tonidero: I think we should avoid using sealed classes/interfaces since it can easily cause breaking changes :(. I think just making this an abstract class, with each of the types being implementations of that class would be the best. An example of the approach would be the recently added [CustomVariableValue](https://github.com/RevenueCat/purchases-android/blob/ca8c53ef876d385c9b9d76458d87879fb27b2fdf/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/CustomVariableValue.kt#L28) in purchases-android.

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -29,28 +51,214 @@ internal fun UIKitPaywall(
...
+                // Standard path: no purchase logic.
+                val paywallViewController = if (footer) {
+                    RCPaywallFooterViewController(
+                        offering = options.offering?.toIosOffering() as? RCOffering,
+                        displayCloseButton = options.shouldDisplayDismissButton,
+                        shouldBlockTouchEvents = false,
+                        dismissRequestedHandler = dismissRequestedHandler,
+                    )
+                } else {
+                    RCPaywallViewController(
```
tonidero: Hmm I wonder if we can unify this to still use the newer `createPaywallViewWithParams`?

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/SamplePurchaseLogic.kt
rickvdl: Was also in doubt about that in the other PRs, but I think that way it's not really clear that this is a different codepath when testing? What do you think? 🤔

revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt
rickvdl: That's a very good point, will see how to address that

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
rickvdl: Added the `displayCloseButton` param to PHC, and just implemented that now :)

revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt
rickvdl: Great idea! Added [here](https://github.com/RevenueCat/purchases-kmp/blob/107296c4809f0c269918035dd7b0c324d8605f6b/revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt#L122-L140)

revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt
rickvdl: Refactored it to use the same API as Android, is this what you had in mind? :)

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/SamplePurchaseLogic.kt
tonidero: Well... It's true that in terms of the backend it would be pretty similar to a "normal" purchase, other than a different `PurchasesAreCompletedBy` value... but this way we can also test real scenarios of a purchase happening in here. I did catch a few bugs in android by following that process, and feels like a more useful test IMO.

revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt
tonidero: Yeah that's great! TBH, I was thinking to keep only the Package as the parameter for now, and add the others later, but I guess we can add them now as well :)

revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt
```diff
@@ -0,0 +1,140 @@
...
+    public val replacementMode: ReplacementMode?,
+    public val subscriptionOption: SubscriptionOption?,
+) {
+    /**
+     * Builder for creating [PaywallPurchaseLogicParams].
+     *
+     * @param rcPackage The package representing the in-app product that the user intends
+     * to purchase.
+     */
+    public class Builder(private val rcPackage: Package) {
```
tonidero: Actually, we might not need the builder... Or at least it should be internal (This is an issue in purchases-android). These parameters are just something that the developer will receive to perform the purchase on their side, but they shouldn't need to create instances for it. 

So IMO, I would remove the builder, and just keep the constructor internal.

apiTester/src/commonMain/kotlin/com/revenuecat/purchases/kmp/apitester/PaywallPurchaseLogicAPI.kt
```diff
@@ -0,0 +1,57 @@
...
+
+
+@Suppress("unused", "UNUSED_VARIABLE", "RedundantExplicitType")
+private class PaywallPurchaseLogicAPI {
+    suspend fun check(offering: Offering?, rcPackage: Package, customerInfo: CustomerInfo) {
+        val purchaseLogic = object : PaywallPurchaseLogic {
+            override suspend fun performPurchase(params: PaywallPurchaseLogicParams): PurchaseLogicResult {
+                val pkg: Package = params.rcPackage
+                val oldProductId: String? = params.oldProductId
+                val replacementMode: com.revenuecat.purchases.kmp.models.ReplacementMode? = params.replacementMode
```
JayShortway: `ReplacementMode` can be imported. (Sorry, just spotted that haha)

composeApp/src/commonMain/kotlin/com/revenuecat/purchases/kmp/sample/SamplePurchaseLogic.kt
rickvdl: Good point, implemented that :)

revenuecatui/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallPurchaseLogic.kt
rickvdl: Sure, just removed them to keep it surface small for now :)

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -19,38 +38,236 @@ internal fun UIKitPaywall(
...
+            resultString = HybridPurchaseLogicBridge.resultError(),
+            errorMessage = result.errorDetails?.let {
+                it.underlyingErrorMessage ?: it.code.description
+            },
+        )
+    }
+}
+
+private suspend fun fetchCustomerInfo(): CustomerInfo =
+    kotlinx.coroutines.suspendCancellableCoroutine { continuation ->
```
tonidero: Could import :)

PR #723, opened 2026-03-05
Migrates iOS from PHC to `kn-core` and `kn-ui`

JayShortway:
## Motivation

## Changes
This PR starts using the `kn-core` and `kn-ui` facade modules added previously. 

## Reviewing
Apologies for the size of this PR, but anything smaller and CI wouldn't pass. The commits are fairly logical units, so you could review commit-by-commit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,13 +37,6 @@ commands:
           name: Install Android SDK platform tools
           command: sdkmanager "platform-tools"
 
-  install-cocoapods-on-macos:
-    steps:
-      - run:
-          name: Install CocoaPods
-          command: bundle exec pod install --repo-update
-          working_directory: iosApp
-
   save-kotlin-native-compiler-to-cache:
     steps:
       - run:
@@ -81,7 +74,7 @@ commands:
       - persist_to_workspace:
           name: Save incremental Gradle build to workspace
           root: ./
-          paths: 
+          paths:
             - .gradle
             # This doesn't support directory glob patterns (**), unfortunately.
             - build
@@ -232,7 +225,6 @@ jobs:
       - attach-workspace-at-working-directory
       - revenuecat/install-gem-mac-dependencies:
           cache-version: v1
-      - install-cocoapods-on-macos
       - run:
           name: Build libraries for iOS
           command: |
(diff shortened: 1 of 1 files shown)


core/core.podspec
tonidero, line 1: 🔥

core/build.gradle.kts
```diff
@@ -21,6 +20,7 @@ kotlin {
             implementation(projects.mappings)
         }
         iosMain.dependencies {
+            implementation(projects.knCore)
```
tonidero, line 23: Sooo much cleaner 🤩

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/AdTracker.ios.kt
```diff
@@ -1,123 +1,62 @@
...
-            data.networkName?.let { put(KEY_NETWORK_NAME, it) }
-            put(KEY_MEDIATOR_NAME, data.mediatorName.value)
-            put(KEY_AD_FORMAT, data.adFormat.value)
-            put(KEY_PLACEMENT, data.placement)
-            put(KEY_AD_UNIT_ID, data.adUnitId)
-            put(KEY_IMPRESSION_ID, data.impressionId)
-        }
-
-        RCCommonFunctionality.trackAdDisplayed(adData)
+        iosAdTracker.trackAdDisplayed(data.toIos())
```
tonidero, line 24: Thanks for cleaning this up! 🙇

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallOptionsKtx.kt
```diff
@@ -29,22 +29,22 @@ internal class IosPaywallDelegate(
     @Suppress("CAST_NEVER_SUCCEEDS")
     override fun paywallViewController(
         controller: RCPaywallViewController,
-        didStartPurchaseWithPackage: RCPackage
+        didStartPurchaseWithPackage: RCPackageFromKnUi,
```
tonidero, line 32: These are interesting aliases 😅 I wonder if we should just prefix these with KN? Like `KNPackage` or `KNCustomerInfo`? Not sure since it seems we need to use the versions for the UI library right? So it might be more confusing 😅  NABD though.

JayShortway: Yea this is something that I'd like to someday fix, but is not really the priority as things are working as-is. The problem is that `RevenueCatUI` references a type (`RCPackage`) from `RevenueCat`, but Kotlin's cinterop doesn't recognize them as the same type. So it generates a Kotlin binding for `RCPackage` in `kn-core` **and** in `kn-ui`. 

We might be able to fix this in our Gradle plugin by giving cinterop the right inputs, but it requires some investigation.

Issue #759, opened 2026-03-20
Undefined symbol: _kniprot_swiftPMImport_com_revenuecat_purchases_kn_core1_RCPurchasesDelegate


MessiasLima:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read
  the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched
  for [existing Github issues](https://github.com/RevenueCat/purchases-kmp/issues)

**Describe the bug**
After the configuration of Revenue Cat following the documentation, im getting the error:

```
Undefined symbol: _kniprot_swiftPMImport_com_revenuecat_purchases_kn_core1_RCPurchasesDelegate
```

1. Environment
    1. Platform: iOS (Compose Multiplatform)
    2. SDK version: 2.9.0+17.52.0
    3. OS version: ios 26.2
    4. IDE (e.g. Android Studio, Xcode, Fleet): Android Studio and Xcode
    5. IDE version: Android Studio 2025.3.2 | Xcode 26.3
    6. Device and/or emulator/simulator:
        - [ ] Device
        - [x] Emulator/simulator
    7. Environment:
        - [x] Closed testing / Sandbox
        - [ ] TestFlight
        - [ ] Production
    8. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
I have just installed the RevenueCat dependencies.
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to
   have context, eg. stackoverflow, etc.)

**Additional context**

<img width="267" height="133" alt="Image" src="https://github.com/user-attachments/assets/77df9029-f7dc-418c-8a26-9cbeefa55233" />

```toml
# libs.version.toml
revenuecat = "2.9.0+17.52.0"

```

<img width="627" height="239" alt="Image" src="https://github.com/user-attachments/assets/5abc9e8e-a78a-4116-94e4-78497c445a24" />

<img width="613" height="177" alt="Image" src="https://github.com/user-attachments/assets/8fe4e982-6b8c-4498-ad83-8fd869154f5f" />

[...]

PR #762, opened 2026-03-23
Fix Xcode 26.3 integration test to reproduce actual user linker scenario

facumenzella:
## Summary

- Adds a new `build-ios-app-xcode26` job that reproduces the exact user scenario from #759
- `build-libraries-ios` (Xcode 16) now runs `publishToMavenLocal` and persists `~/.m2` to the workspace
- New standalone `ciIntegrationTests/` Gradle project consumes `purchases-kmp-core` from mavenLocal and installs `PurchasesHybridCommon` via CocoaPods
- `build-ios-app-xcode26` links that pre-built binary with Xcode 26.3's linker — the exact path that triggers the error for end-users
- Renames `build-sample-ios` → `build-sample-ios-xcode16` for clarity

## Motivation

Issue #759 reports an undefined linker symbol (`_kniprot_swiftPMImport_com_revenuecat_purchases_kn_core1_RCPurchasesDelegate`) on Xcode 26.2/26.3.

The key detail: end-users don't build the KMP library — they link against a **pre-built binary** (published with Xcode 16.x). Running `xcodebuild` against `iosApp` wouldn't reproduce this because it triggers the `Compile Kotlin Framework` build phase, which recompiles purchases-kmp with Xcode 26.3's toolchain.

The correct reproduction is:
1. Build the KMP library with Xcode 16 → `publishToMavenLocal`
2. A fresh consumer project pulls from mavenLocal and links with Xcode 26.3

`ciIntegrationTests/` is that consumer project.

## Test plan

- [ ] CI passes on Xcode 16.4 jobs (no regression)
- [ ] `build-ios-app-xcode26` either reproduces the linker error from #759 or confirms a clean link with Xcode 26.3

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -313,7 +313,7 @@ jobs:
       - save-gradle-user-home-directory-to-cache
       - save-incremental-gradle-build-to-workspace
 
-  build-sample-ios-xcode16:
+  build-sample-ios:
     executor: xcode16
     steps:
       - checkout
@@ -525,7 +525,7 @@ workflows:
       - build-libraries-ios
       - build-sample-android:
           requires: [build-libraries-android]
-      - build-sample-ios-xcode16:
+      - build-sample-ios:
           requires: [build-libraries-ios]
       - build-ios-app-xcode26:
           requires: [build-libraries-ios]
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -284,6 +309,42 @@ jobs:
       - save-kotlin-native-compiler-to-cache
       - save-incremental-gradle-build-to-workspace
 
+  build-ios-app-xcode26:
```
JayShortway, line 339:
```suggestion
  build-sample-ios-xcode26:
```

PR #767, opened 2026-03-25
Migrates more iOS features from PHC to `kn-core` and `kn-ui`

JayShortway:
## Motivation

## Description
Follow-up to https://github.com/RevenueCat/purchases-kmp/pull/723 which migrates the new features added after that PR got merged: 
- presenting code redemption sheets
- tracking custom paywall impressions
- custom Paywall variables
- custom PurchaseLogic

## About the CI failure
This is ready for review. CI will succeed when the new version of purchases-ios gets released which includes https://github.com/RevenueCat/purchases-ios/pull/6507.

### DIFF
--- a/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
+++ b/revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
@@ -113,10 +113,6 @@ private class PurchaseHandler(
                 error = (result as? PurchaseLogicResult.Error)?.errorDetails?.toNSError()
                 success = result is PurchaseLogicResult.Success
             } catch (e: CancellationException) {
-                error = PurchasesError(
-                    code = PurchasesErrorCode.UnknownError,
-                    underlyingErrorMessage = "Restore cancelled"
-                ).toNSError()
                 throw e
             } catch (t: Throwable) {
                 error = PurchasesError(
@@ -143,10 +139,6 @@ private class PurchaseHandler(
                 error = (result as? PurchaseLogicResult.Error)?.errorDetails?.toNSError()
                 userCancelled = result is PurchaseLogicResult.Cancellation
             } catch (e: CancellationException) {
-                error = PurchasesError(
-                    code = PurchasesErrorCode.PurchaseCancelledError,
-                    underlyingErrorMessage = "Purchase cancelled"
-                ).toNSError()
                 userCancelled = true
                 throw e
             } catch (t: Throwable) {


revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -124,168 +74,72 @@ internal fun UIKitPaywall(
...
+    variables.forEach { (key, value) ->
+        when (value) {
+            is CustomVariableValue.String -> setCustomVariable(value = value.value, forKey = key)
+            is CustomVariableValue.Number -> setCustomVariableNumber(
+                value = value.value,
+                forKey = key
+            )
+            is CustomVariableValue.Boolean -> setCustomVariableBool(
+                value = value.value,
+                forKey = key
```
rickvdl, line 88: nit: we previously had an else case here that would log for unsupported variable types

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -124,168 +74,72 @@ internal fun UIKitPaywall(
...
+        coroutineScope.launch {
+            var complete = false
+            try {
+                val result = purchaseLogic.performRestore(getCustomerInfo())
+                val error = (result as? PurchaseLogicResult.Error)?.errorDetails?.toNSError()
+                val success = result is PurchaseLogicResult.Success
+                completion(success, error)
+                complete = true
+            } finally {
+                if (!complete) completion(!isActive, null)
```
rickvdl: I believe the error (second parameter) here is required for the native side to treat it as such. So I think in theory `purchaseLogic.performRestore(getCustomerInfo())` could throw and end up in that codepath? Same on line 135 :)

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -37,85 +33,39 @@ internal fun UIKitPaywall(
...
-                RCPaywallFooterViewController(
-                    offering = options.offering?.toIosOffering() as? RCOffering,
-                    displayCloseButton = options.shouldDisplayDismissButton,
-                    shouldBlockTouchEvents = false,
-                    dismissRequestedHandler = dismissRequestedHandler,
-                ).apply { setDelegate(footerDelegate) }
-                    .also { layoutViewControllerState.setViewController(it) }
-            }
+            val paywallViewController = if (footer) RCPaywallFooterViewController(
+                offering = options.offering?.toIosOffering() as? RCOfferingFromKnUi,
```
rickvdl, line 53: We previously did not add the custom purchase logic handling here to the footer, since we considered it in "maintenance mode". Do we want to explicitly add it here?

JayShortway: Oh hmm right. I don't have a very strong opinion on this. It's supported by https://github.com/RevenueCat/purchases-ios/pull/6507 so I figured we can wire it up as it's free at this point.

rickvdl: Same, fine either way by me, cc @tonidero for your input :)

tonidero: Right... I don't have a strong opinion either. Mostly my thought was that, if we don't, it would be a way to try to move more people to the newer Paywalls...But having said that, if it's already done, I'm not opposed to keeping it

kn-core/src/swift/Sources/AdditionalSwift/AppleApiAvailability.swift
```diff
@@ -35,6 +35,22 @@ public class AppleApiAvailability: NSObject {
         #endif
     }
 
+    /// Determines if the `presentCodeRedemptionSheet` API is available on the current device.
+    ///
+    /// - Returns: `true` if `presentCodeRedemptionSheet` is available, `false` otherwise.
+    @objc
+    public func isCodeRedemptionSheetAPIAvailable() -> Bool {
```
tonidero, line 42: I wonder if these are something that we should expose from purchases-ios?

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
```diff
@@ -124,168 +74,72 @@ internal fun UIKitPaywall(
...
+                complete = true
+            } finally {
+                if (!complete) completion(!isActive, null)
+            }
         }
     }
+
+    private fun PurchasesError.toNSError() = NSError(
+        code = code.code.toLong(),
+        domain = "com.revenuecat.purchases",
```
tonidero: Nitpick but would it be worth customizing the domain a bit:
```suggestion
        domain = "com.revenuecat.purchases.kmp",
```

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
JayShortway: Oh right, I somehow thought these were still sealed (and thus exhaustive). Else case added: https://github.com/RevenueCat/purchases-kmp/pull/767/commits/0e266ca79ba3d3dc22b5291bc9582e8f28638933.

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
JayShortway: Good catch! I think this should do it: https://github.com/RevenueCat/purchases-kmp/pull/767/commits/85b3f999133753d5d84ea78938462d34615f4e04

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
JayShortway: I can agree to that! https://github.com/RevenueCat/purchases-kmp/pull/767/commits/ac93c3052692653ffaa593044a702cd940e24a32

kn-core/src/swift/Sources/AdditionalSwift/AppleApiAvailability.swift
JayShortway: It's currently in [purchases-hybrid-common](https://github.com/RevenueCat/purchases-hybrid-common/blob/bdbf9b52824436a0e7166f9b46082d4a7febadf6/ios/PurchasesHybridCommon/PurchasesHybridCommon/IOSAPIAvailabilityChecker.swift). In purchases-ios we could just use Swift's built-in annotations and macros right?

tonidero: Right indeed. But wondering if it would be worth exposing to customers, like an API to check whether this functionality would work... But well, I think keeping it here is ok for now!

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
JayShortway: Alright, in that case I'll go with the path of least resistance, which is leaving it in. It's also supported on Android, so it could be confusing if we don't support it on iOS.

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
rickvdl: Hm if I'm reading it correctly we're now returning an error on cancellation, which I don't think is expected by the native side. If an error is returned it's shown to the user I believe. IIRC the cancel case should return false with a null value for the error. Looking at https://github.com/RevenueCat/purchases-ios/blob/2c29ffd5b5f87a04268a1f8ec912ea8921628bf0/RevenueCatUI/Purchasing/PurchaseHandler.swift#L338-L342

kn-core/src/swift/Sources/AdditionalSwift/AppleApiAvailability.swift
ajpallares: Even if we exposed such API in purchases-ios, the Swift compiler would still force developers to add these `#if` and `if #available` checks before they called the code redemption sheet API... so I think having `isCodeRedemptionSheetAPIAvailable()` wouldn't be useful

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
ajpallares: Agree with this!

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/UIKitPaywall.kt
JayShortway: At least one of us is paying attention 😓 Third time's the charm? https://github.com/RevenueCat/purchases-kmp/pull/767/commits/ee36d4982007395ceade8b8365b3f2c36728a244 (Severe Friday brain, so maybe not.)

rickvdl: I _think_ that looks good, but it's the same day for me 😉. We could set `userCancelled = true` for the restore cancellation as well but I don't think it's required.

JayShortway: Restore has a `success` boolean and optional `error`. Purchase has a `userCancelled` boolean and optional `error`. 😅

rickvdl: Aha 😅

PR #770, opened 2026-03-25
Fixes release automations

JayShortway:
## Motivation

## Description
This PR fixes release automations after migrating both the iOS and Android targets away from PHC. This mainly involves updating the `VERSIONS.md` file correctly upon release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,14 +114,8 @@ commands:
 parameters:
   action:
     type: enum
-    enum: [ build, upgrade-hybrid-common, bump ]
+    enum: [ build, bump ]
     default: build
-  automatic:
-    type: boolean
-    default: false
-  version:
-    type: string
-    default: ''
 
 executors:
   android202409:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -436,15 +413,6 @@ jobs:
           command: bundle exec fastlane github_release_current_version
 
 workflows:
-  on-action-upgrade-hybrid-common:
```
tonidero, line 439: We will need to change PHC to avoid calling this non-existing workflow after merging this to main

fastlane/Fastfile
```diff
@@ -179,40 +116,67 @@ lane :github_release do |options|
...
+  version = File.read("../upstream/purchases-ios/.version").strip
+  UI.user_error!("iOS version not found in upstream/purchases-ios/.version") if version.nil? || version.empty?
+  version
+end
 
-  def check_no_git_tag_exists(version_number)
-    if git_tag_exists(tag: version_number, remote: true, remote_name: 'origin')
-      raise "git tag with version #{version_number} already exists!"
-    end
+def get_billing_client_version(android_version)
```
tonidero, line 139: Wonder if we should extract this to the fastlane plugin so it can be shared with PHC... Not a blocker though :)

fastlane/Fastfile
```diff
@@ -179,40 +116,67 @@ lane :github_release do |options|
...
+  UI.message("iOS version: #{ios_version}")
+  UI.message("Play Billing Library version: #{billing_version}")
+
+  versions_path = File.join('..', PATH_VERSIONS)
+  lines = File.readlines(versions_path)
+  new_line = [
+    sdk_version,
+    "[#{ios_version}](https://github.com/RevenueCat/purchases-ios/releases/tag/#{ios_version})",
+    "[#{android_version}](https://github.com/RevenueCat/purchases-android/releases/tag/#{android_version})",
+    "N/A",
```
tonidero, line 168: Hmm good to leave this like this for now I think. Once we've merge this and some time passes, it might be worth removing the column entirely.

.circleci/config.yml
ajpallares: Great point! I think this should be enough https://github.com/RevenueCat/purchases-hybrid-common/pull/1583

fastlane/Fastfile
JayShortway: Hmm yea, wouldn't that require us to make it a lane instead of a function?

tonidero: Indeed.... so yeah, I guess it's not that complex and not that reused just now that we need to do this I think

PR #771, opened 2026-03-25
Updates Kotlin to 2.3.20 and Gradle to 9.4.1

JayShortway:
## Motivation
We're doing a major. So I figured while we're at it, we can update the Kotlin version.


build.gradle.kts
```diff
@@ -5,7 +5,6 @@ import com.vanniktech.maven.publish.MavenPublishPlugin
 import com.vanniktech.maven.publish.SonatypeHost
 import dev.iurysouza.modulegraph.Orientation
 import io.gitlab.arturbosch.detekt.Detekt
-import org.gradle.configurationcache.extensions.capitalized
```
tonidero, line 8: Just curious, was this function removed?

JayShortway: Yea, the error said something along the lines of "this was never intended to be a public API" 😅, with an automatic fix (which I turned into the extension function below).

revenuecatui/api/revenuecatui.api
```diff
@@ -57,13 +57,13 @@ public final class com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallKt {
...
-	public abstract fun onRestoreCompleted (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;)V
-	public abstract fun onRestoreError (Lcom/revenuecat/purchases/kmp/models/PurchasesError;)V
-	public abstract fun onRestoreStarted ()V
+	public fun onPurchaseCancelled ()V
+	public fun onPurchaseCompleted (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;Lcom/revenuecat/purchases/kmp/models/StoreTransaction;)V
+	public fun onPurchaseError (Lcom/revenuecat/purchases/kmp/models/PurchasesError;)V
+	public fun onPurchaseStarted (Lcom/revenuecat/purchases/kmp/models/Package;)V
+	public fun onRestoreCompleted (Lcom/revenuecat/purchases/kmp/models/CustomerInfo;)V
+	public fun onRestoreError (Lcom/revenuecat/purchases/kmp/models/PurchasesError;)V
+	public fun onRestoreStarted ()V
```
tonidero, line 66: Hmm interesting... was this leftover from previous changes? Or is it that the compiled bytecode is different in new kotlin version? Just to make sure why it changed to be not abstract now 😅

revenuecatui/api/revenuecatui.klib.api
```diff
@@ -130,12 +130,10 @@ final class com.revenuecat.purchases.kmp.ui.revenuecatui/PaywallPurchaseLogicPar
         final fun <get-rcPackage>(): com.revenuecat.purchases.kmp.models/Package // com.revenuecat.purchases.kmp.ui.revenuecatui/PaywallPurchaseLogicParams.rcPackage.<get-rcPackage>|<get-rcPackage>(){}[0]
 }
 
-final val com.revenuecat.purchases.kmp.ui.revenuecatui.modifier/com_revenuecat_purchases_kmp_ui_revenuecatui_modifier_LayoutViewControllerState$stableprop // com.revenuecat.purchases.kmp.ui.revenuecatui.modifier/com_revenuecat_purchases_kmp_ui_revenuecatui_modifier_LayoutViewControllerState$stableprop|#static{}com_revenuecat_purchases_kmp_ui_revenuecatui_modifier_LayoutViewControllerState$stableprop[0]
```
tonidero, line 133: Interesting that all these disappeared in this PR... but yeah maybe the kotlin version bump changed the bytecode and these aren't needed anymore?

revenuecatui/api/revenuecatui.api
JayShortway: Yea my assumption is that the api format stopped including the `abstract` keyword in the new Kotlin version?

revenuecatui/api/revenuecatui.klib.api
JayShortway: I think it's something like that yea.

PR #779, opened 2026-04-01
Removes dead code

JayShortway:
## Description
Removes some code that was only used in tests.


tonidero (approved):
Oh hmm interesting. Thanks for removing this!

PR #780, opened 2026-04-01
Fixes caching of the Swift build tasks

JayShortway:
## Issue
Both a Gradle Sync in Android Studio and any Gradle compilation task would compile the Swift source files, without reusing the output of each other. This only really affects local development, as Gradle Sync is not a thing that happens on CI. 

## Fix
The fix in this PR is to make all Gradle tasks responsible for building Swift a `@CacheableTask`. This has the effect that configuration cache misses no longer retrigger these tasks if their inputs (i.e. Swift source files) haven't changed.


tonidero (approved):
Ohh interesting… I think it looks good!

PR #785, opened 2026-04-02
Adds `iosApp` run config

JayShortway:
## Description
The [Kotlin Multiplatform plugin for Android Studio](https://plugins.jetbrains.com/plugin/14936-kotlin-multiplatform) enables building the iOS tester app from Android Studio:  
  
afbeelding

However, this wasn't working consistently. On top of that we've made some changes to the tester app due to the removal of CocoaPods, which might have broken it further. 

This PR adds the generated run configuration (`.run/iosApp.run.xml`) to the repo, including some other files on which this functionality depends. We'll have to see how stable these files are. In any case this should hopefully make this work much more consistently for everyone.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.iml
 .gradle
 **/.idea/*
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
-*.iml
 .gradle
-.idea
+**/.idea/*
 .DS_Store
 build
 captures
@@ -21,4 +20,8 @@ vendor/
 .bundle/
 
 # Share IssueNavigationConfiguration
-!.idea/vcs.xml
+!/.idea/vcs.xml
+# Ensure iosApp run configuration works
+!/.idea/xcode.xml
+!/.idea/purchases-kmp.iosApp.iml
+!/.idea/modules.xml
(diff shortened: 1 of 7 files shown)


.run/iosApp.run.xml
```diff
@@ -0,0 +1,8 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="iosApp" type="AppleRunConfiguration" factoryName="Application" REDIRECT_INPUT="false" ELEVATE="false" USE_EXTERNAL_CONSOLE="false" EMULATE_TERMINAL="false" PASS_PARENT_ENVS_2="true" PROJECT_NAME="iosApp" TARGET_NAME="iosApp" CONFIG_NAME="Debug" IS_LOCATION_SIMULATION_SUPPORTED="true" SCHEME_NAME="iosApp" IS_LOCATION_SIMULATION_ALLOWED="true" LOCATION_SCENARIO_ID="com.apple.dt.IDEFoundation.CurrentLocationScenarioIdentifier" LOCATION_SCENARIO_TYPE="1" APPLICATION_LANGUAGE="IDELaunchSchemeLanguageUseSystemLanguage" APPLICATION_REGION="" DEVELOPMENT_TEAM="8SXR2327BM" RUN_TARGET_PROJECT_NAME="iosApp" RUN_TARGET_NAME="iosApp" MAKE_ACTIVE="TRUE" SHOULD_DEBUG_EXTENSIONS="false">
```
rickvdl, line 2: Are (some of these) settings configurable in the IDE? Mostly asking about `DEVELOPMENT_TEAM` since that is required to be changed when running on device because of codesigning

.gitignore
```diff
@@ -1,6 +1,5 @@
-*.iml
```
rickvdl, line 1: Do we no longer want to ignore these globally? I think currently only `.idea/purchases-kmp.iosApp.iml` is added

rickvdl (approved):
Not the easiest diff but I think it makes sense if it (possibly) improves reliability :) Two small questions

tonidero (approved):
Just existing comments. Thanks for making this easier!! 🙇

.run/iosApp.run.xml
JayShortway: Yes, you can modify e.g. the development team in the IDE by editing this run configuration (in the UI), which would in turn modify this file. It's a bit of a tradeoff. My idea is to see how this goes. If we constantly have accidental changes committed to this file, we can think about alternatives?

.gitignore
JayShortway: Good one, added back here: https://github.com/RevenueCat/purchases-kmp/pull/785/commits/9c746f8e6056fa56f4b6cb20777404bbcc27a4c0. They're all in the `.idea` folder normally, but no need to encode that assumption in the `.gitignore`.

.run/iosApp.run.xml
rickvdl: Ah that makes sense. Yes indeed, let's see how it turns out :)

PR #792, opened 2026-04-07
Fix flaky build-libraries-ios by serializing pod installs

ajpallares:
## Summary
- Fixes the intermittent `build-libraries-ios` CI failure caused by a CocoaPods cache race condition
- Gradle runs `podInstallSyntheticIos` for `core`, `models`, `mappings`, and `revenuecatui` in parallel. All four hit the shared `~/Library/Caches/CocoaPods/Pods/` directory concurrently, causing `Errno::ENOENT` errors during `cp_r` (e.g. `No such file or directory @ rb_file_s_lstat`)
- Adds a dedicated step that runs the pod install tasks with `--no-parallel` before the main build. The compile tasks still run fully parallel since pods are already installed (UP-TO-DATE)

## Test plan
- [x] Verify the `build-libraries-ios` job passes on this branch's CI run. Stress tested [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-kmp/2856). 20 runs, 0 failures.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,7 +125,7 @@ commands:
 parameters:
   action:
     type: enum
-    enum: [ build, upgrade-hybrid-common, bump ]
+    enum: [ build, upgrade-hybrid-common, bump, stress-test-ios ]
     default: build
   automatic:
     type: boolean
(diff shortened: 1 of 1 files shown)


tonidero (approved):
Ohhh thank you so much for fixing this 🙇

PR #798, opened 2026-04-08
Updates Fastlane `README.md`

JayShortway:
## Description
As the title says. Can't release locally without this because it makes the git status dirty 🙈

### DIFF
--- a/fastlane/README.md
+++ b/fastlane/README.md
@@ -29,14 +29,6 @@ Bumps version, edits changelog, and creates pull request
 
 Automatically determines next version, bumps it, edits changelog, and creates pull request
 
-### update_hybrid_common
-
-```sh
-[bundle exec] fastlane update_hybrid_common
-```
-
-Update purchases-hybrid-common dependency
-
 ### prepare_next_snapshot_version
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -292,9 +292,9 @@ GEM
     naturally (2.3.0)
     netrc (0.11.0)
     nkf (0.2.0)
-    nokogiri (1.19.2-arm64-darwin)
+    nokogiri (1.18.10-arm64-darwin)
       racc (~> 1.4)
-    nokogiri (1.19.2-x86_64-linux-gnu)
+    nokogiri (1.18.10-x86_64-linux-gnu)
```
tonidero: Hmm this might be due to a mismatch in the ruby version... But I think it probably will work fine on CI, so happy to leave this as is!

tonidero (approved):

Gemfile.lock
JayShortway: Ahh good catch, you're right! Reverted.

ajpallares (approved):

PR #810, opened 2026-04-15
Removes unnecessary CI steps

JayShortway:
## Description
- `pin-ruby-version` is no longer needed, as that was a fix for a CocoaPods issue.
- There's no point in writing to the workspace when there's no subsequent job using that workspace.

Both of these steps were significant. Removing them shaves of 2+ minutes of a typical workflow run, meaning it now gets close to (sometimes under) 10 minutes.


ajpallares (approved):
Makes sense!

PR #819, opened 2026-04-20
Enables Gradle metadata and wrapper verification

JayShortway:
## Description
This enables Gradle dependency verification and wrapper verification.

## How dependency verification is implemented 
1. Added a `build-settings` plugin to pin every dependency to a single repository. More info below.
2. Added a minimal `gradle/verification-metadata.xml` file with the following contents:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<verification-metadata xmlns="https://schema.gradle.org/dependency-verification" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://schema.gradle.org/dependency-verification https://schema.gradle.org/dependency-verification/dependency-verification-1.3.xsd">
   <configuration>
      <verify-metadata>true</verify-metadata>
      <verify-signatures>true</verify-signatures>
      <keyring-format>armored</keyring-format>
      <key-servers enabled="false"/>
   </configuration>
</verification-metadata>
```
3. Bootstrapped the trusted keys and checksums by running:
```shell
./gradlew --refresh-dependencies \
  --refresh-keys \
  --write-verification-metadata pgp,sha256 \
  --export-keys 
```

## How dependency verification works
When Gradle downloads a dependency, it will check if it's signed. 
- If it's signed, it will validate the key with the known keys in `gradle/verification-keyring.keys`.
- If it's not signed, it will validate the checksum with the known checksum in `gradle/verification-metadata.xml`.

## The build-setings plugin
Many dependencies exist on multiple repositories. Annoyingly, it's common for the dependencies to be slightly different on each repository. 
- Sometimes they're signed on one repository but not on others.
- Sometimes they're not byte-for-byte identical, meaning the checksums differ.

For this reason I added configuration that pins each dependency to a single repository. This makes the behavior deterministic, at the cost of friction when adding a new dependency. I moved this to a settings plugin so we don't have to duplicate this configuration between the main build and the existing `build-logic` plugin.

### Updating a dependency version
1. Update the version in `gradle/libs.versions.toml`.
2. Run the command from step 3 above.

### Adding a new dependency
1. Add the dependency to `gradle/libs.versions.toml` and any `build.gradle.kts` files.
2. Pin it to a repository in the `build-settings` plugin.
3. Run the command from step 3 above.


gradle/verification-metadata.xml
```diff
@@ -0,0 +1,1032 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<verification-metadata xmlns="https://schema.gradle.org/dependency-verification" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://schema.gradle.org/dependency-verification https://schema.gradle.org/dependency-verification/dependency-verification-1.3.xsd">
+   <configuration>
+      <verify-metadata>true</verify-metadata>
+      <verify-signatures>true</verify-signatures>
+      <keyring-format>armored</keyring-format>
```
JayShortway, line 6: This format is human readable, which is nicer to review in PRs and avoids having to commit a binary file.

gradle/verification-metadata.xml
```diff
@@ -0,0 +1,1032 @@
...
+      <key-servers enabled="false"/>
```
JayShortway, line 7: This means Gradle won't reach out to key servers to validate signing keys. It will only use the keys in `gradle/verification-keyring.keys`. That helps with reproducibility and avoids making calls to key servers in CI.

gradle/libs.versions.toml
```diff
@@ -43,7 +43,7 @@ arturbosch-detekt = { id = "io.gitlab.arturbosch.detekt", version.ref = "detekt"
 android-application = { id = "com.android.application", version.ref = "agp" }
 android-library = { id = "com.android.library", version.ref = "agp" }
 codingfeline-buildkonfig = { id = "com.codingfeline.buildkonfig", version = "0.15.1" }
-iurysouza-modulegraph = { id = "dev.iurysouza.modulegraph", version = "0.10.0" }
+iurysouza-modulegraph = { id = "dev.iurysouza.modulegraph", version = "0.13.0" }
```
JayShortway, line 46: This was needed because version `0.10.0` isn't compatible with Gradle 9 (which we've been using for a while). Not sure why this wasn't a problem before.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionSha256Sum=2ab2958f2a1e51120c326cad6f385153bb11ee93b3c216c5fccebfdfbb7ec6cb
```
JayShortway, line 3: This is all that's needed to ensure Gradle itself is validated.

rickvdl (commented):
Great improvement! Changes make sense to me but I'll leave approval to someone with a bit more knowledge on Android tooling 🧰

build-settings/src/main/kotlin/revenuecat-repositories.settings.gradle.kts
```diff
@@ -0,0 +1,133 @@
...
+    includeGroup("org.jsoup")
+    includeGroup("org.jspecify")
+    includeGroupByRegex("org\\.junit($|\\..*)")
+    includeGroup("org.jvnet.staxex")
+    includeGroup("org.opentest4j")
+    includeGroupByRegex("org\\.ow2($|\\..*)")
+    includeGroup("org.slf4j")
+    includeGroup("org.snakeyaml")
+    includeGroup("org.sonatype.oss")
+    includeGroup("org.tensorflow")
```
tonidero, line 87: Would it be worth adding instructions on how to add/update dependencies and plugins in AGENTS.md? Maybe even in the README.md... seems like something we might forget at some point.

tonidero (approved):
I think it makes sense. Thanks for doing this! It's true that it might be a bit more work... but probably good to stay safe :)

JayShortway:
@RevenueCat/coresdk I'm going to close this for now. It can serve as a reference in case we want to revisit this later.

Reason being that this adds too much friction. I realized that bootstrapping verification metadata (e.g. `--refresh-keys  --write-verification-metadata pgp,sha256`) only adds the metadata (checksums and keys) for the Gradle task you're performing. I added it while running the `tasks` task. Now when I run `build`, it might use other transitive dependencies not yet in `verification-metadata.xml` and fail. When I add those, it might fail on `publish` next. Right now I tried a Gradle sync in Android Studio and it failed. 

So every time we change a dependency we need to exhaustively run all (important) Gradle tasks to bootstrap the metadata. That's extremely prohibitive. In my opinion we need to come up with a good workflow that balances convenience with security before adopting metadata verification.

PR #825, opened 2026-04-27
Add maestro E2E test for purchase through paywall

ajpallares:
### Motivation

Adds Maestro test flows for the KMP E2E test app.

### Description

* `purchase_through_paywall.yaml`: Launches the app, verifies no entitlements, presents the paywall, makes a purchase, and verifies the "pro" entitlement
* `confirm_purchase.yaml`: Reusable subflow for the sandbox purchase confirmation dialog
* `config.yaml`: Workspace configuration for test discovery

**Related PRs:** #824 (E2E test app), #826 (CI job)

PR #826, opened 2026-04-27
Add CircleCI job for maestro E2E tests

ajpallares:
### Motivation

Adds CI jobs to build and run Maestro E2E tests on every PR commit.

### Description

* CircleCI jobs for iOS (macOS) and Android (`machine: android:2025.10.1`) Maestro E2E tests
* Fastlane lanes: `change_maestro_test_app_api_key`, `build_maestro_test_app`, `run_maestro_e2e_tests_ios`, `run_maestro_e2e_tests_android`
* Pre-boots iOS simulator in parallel with build for faster CI
* Uses `run_maestro_e2e_tests` from `fastlane-plugin-revenuecat_internal` for retry logic
* Crash log collection for iOS CI diagnostics
* Xcode build phase handles Kotlin framework + Compose Resources embedding

**Related PRs:** #824 (E2E test app), #825 (Maestro test flows)


Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal
-  revision: 20911d1ac54e7cdcf339bc094b75f928e46d7be0
+  revision: d24ab263633b6a3c7615dcb26d7d6fefb573fecb
```
ajpallares: This is the latest version of the plugin as of today

.circleci/config.yml
```diff
@@ -390,6 +390,100 @@ jobs:
           name: "Create a GitHub release for the current version"
           command: bundle exec fastlane github_release_current_version
 
+  run-maestro-e2e-tests-ios:
+    macos:
+      xcode: 16.4.0
```
ajpallares: I tried using Xcode 26.4.1 in #827 but got this failure:

```
Task :revenuecatui:compileKotlinIosSimulatorArm64 FAILED
Unresolved reference 'RCCustomerInfo'.
Unresolved reference 'RCPackage'.
Unresolved reference 'RCStoreTransaction'.
Unresolved reference 'RCOffering'.
Unresolved reference 'RCOfferingFromKnUi'.
Unresolved reference 'RCPackageFromKnUi'.
```

Seems like we need to do some Xcode 26 compatibility work before we can upgrade to use Xcode 26

tonidero: Sad... We might need to update the version of kotlin https://kotlinlang.org/docs/multiplatform/multiplatform-compatibility-guide.html#version-compatibility. But yeah, that should probably be separate.

.circleci/config.yml
```diff
@@ -390,6 +390,100 @@ jobs:
...
+          avd-name: test-e2e
+          no-window: true
+          restore-gradle-cache-prefix: ""
+          post-emulator-launch-assemble-command: bundle exec fastlane build_maestro_test_app platform:android
+      - save-gradle-user-home-directory-to-cache
+      - save-kotlin-native-compiler-to-cache
+      - revenuecat/install-maestro
+      - run:
+          name: Run Maestro E2E Tests (Android)
+          command: bundle exec fastlane run_maestro_e2e_tests_android
```
tonidero, line 474: I was thinking... Do we need to add a [wait_for_emulator](https://circleci.com/developer/orbs/orb/circleci/android#commands-wait_for_emulator) command before this?

fastlane/Fastfile
```diff
@@ -101,6 +101,63 @@ lane :github_release_current_version do |options|
...
+    new_text: new_api_key,
+    paths_of_files_to_update: [app_kt]
+  )
+  UI.user_error!("API key was not found in #{app_kt} after replacement") unless File.read(app_kt).include?(new_api_key)
+end
+
+desc "Build the Maestro E2E test app for the given platform (platform:ios or platform:android)"
+lane :build_maestro_test_app do |options|
+  platform = options[:platform].to_s.downcase
+  UI.user_error!("Missing or invalid `platform` option (use ios or android)") unless %w[ios android].include?(platform)
```
tonidero: TBH, I wonder if we should just split this into 2 lanes... But NABD either :)

tonidero (approved):
Just some nitpicks but nothing really blocking. Thank you!

fastlane/Fastfile
ajpallares: Right. Good point. There's not any shared work really, so 2 lanes is cleaner, for sure

ajpallares: Done in https://github.com/RevenueCat/purchases-kmp/pull/826/commits/2df4cb327d5a5dbea30c05aef44dd594e818f2b1

.circleci/config.yml
ajpallares: I think it would be redundant here. Since we have a previous `android/start-emulator` that has the `wait_for_emulator` parameter which defaults to `true` ([docs](https://circleci.com/developer/orbs/orb/circleci/android#commands-start_emulator))

tonidero: True, nvm 😅

.circleci/config.yml
ajpallares: I'm working on the Xcode 26 compatibility in https://github.com/RevenueCat/purchases-kmp/pull/828

Do you mean we'll need to update the version of kotlin as well?

If I'm not mistaken, branch 3.0.0-dev is already using kotlin 2.3.20 https://github.com/RevenueCat/purchases-kmp/blob/2aa34d2b12ebe42e82ba8ab7cf1d6d490fb09573/gradle/libs.versions.toml#L13

so we shouldn't need to bump it?

tonidero: Oh right, we're already pointing to that branch. Nvm 😅

PR #827, opened 2026-04-29
Update CI to Xcode 26.4.1 with cinterop compatibility fix

ajpallares:
### Summary

Updates all iOS CI jobs to use Xcode 26.4.1 (from 16.4.0) and fixes the cinterop build-logic to be compatible with the new toolchain.

### Changes

**CI configuration:**
- Renamed `xcode16` executor to `xcode26`, updated to Xcode `26.4.1`
- All iOS jobs now use Xcode 26.4.1: `build-libraries-ios`, `build-sample-ios`, `public-api-tests-ios`, `unit-tests-ios`, `unit-tests-build-logic`, `detekt`, `validate-binary-compatibility`, `publish`

**Build-logic fix for Swift 6.2 cinterop compatibility:**

Two issues caused cinterop to fail with the new toolchain:

1. **Shared SPM scratch directory caused header overwrites**: `-Xswiftc -emit-objc-header-path` applies to ALL swiftc invocations during `swift build`, so a dependency module's header could overwrite the target module's header. Fix: use per-target scratch directories.

2. **Forward-declared dependency types became unresolvable**: the target header uses `@class` forward declarations for dependency types (e.g. `RCCustomerInfo` from RevenueCat), which cinterop turns into unresolvable stubs. Fix: include dependency headers in the target's modulemap so the full `@interface` definitions are available to cinterop.

Dependency `-I` paths are kept in the cinterop compiler options for backward compatibility, as `-fmodules` needs them to resolve forward declarations through the module system.


ajpallares:
Replaced by a clean PR with only the Xcode 26 compatibility changes (no maestro changes).

PR #828, opened 2026-04-29
Add Xcode 26 compatibility

ajpallares:
### Summary

Adds Xcode 26 / Swift 6.2 compatibility to the cinterop build-logic and updates all iOS CI jobs to use Xcode 26.4.1.

### Changes

**CI:**
- Renamed `xcode16` executor to `xcode26`, updated to Xcode `26.4.1`

**Build-logic (cinterop fix for Swift 6.2):**

Two issues caused cinterop to fail with the new toolchain:

1. **Shared SPM scratch directory caused header overwrites**: `-Xswiftc -emit-objc-header-path` applies to ALL swiftc invocations during `swift build`, so a dependency module's header could overwrite the target module's header. Fix: use per-target scratch directories.

2. **Forward-declared dependency types became unresolvable**: the target header uses `@class` forward declarations for dependency types (e.g. `RCCustomerInfo` from RevenueCat), which cinterop turns into unresolvable stubs. Fix: include dependency headers in the target's modulemap so the full `@interface` definitions are available to cinterop.

Dependency `-I` paths are kept in the cinterop compiler options for backward compatibility, as `-fmodules` needs them to resolve forward declarations through the module system.

**Trade-off — duplicate SPM builds:** Per-target scratch dirs mean SPM no longer reuses build artifacts across consumers, so `RevenueCat` gets compiled once for `:kn-core` and again for `:kn-ui`.

**Tests:**
- Added `SwiftCrossProjectDependencyTests` exercising cross-project Swift target dependencies end-to-end
- Auto-derive `evaluationDependsOn` calls in the test DSL from the target dependency graph


JayShortway (approved):
Thanks for fixing this! ❤️

tonidero (approved):
Thanks for doing this!! 🙌

PR #847, opened 2026-05-07
Remove stale PHC build metadata note from RELEASING.md

ajpallares:
## Summary

Since `3.0.0`, the SDK version no longer carries a `+<phc-version>` build metadata suffix (the *Common files version* column in `VERSIONS.md` is now `N/A`, and `update_versions_file` in the Fastfile already writes `N/A` for it).

The manual release instructions in `RELEASING.md` still asked the releaser to consider that build metadata when entering a new version number. That note is stale and could confuse the next person doing a manual release, so this drops it.

This is a follow-up cleanup to RevenueCat/purchases-hybrid-common#1639, which removed the corresponding outgoing trigger from PHC.

## Test plan

* Doc-only change. CI passes.

Made with [Cursor](https://cursor.com)


tonidero (approved):
Thank you!!

PR #848, opened 2026-05-07
Add subscriber attributes testing screen to tester app

vegaro:
### Motivation

The tester app had no way to exercise subscriber attribute setter functions. This screen makes it easy to set or clear any of the 21 subscriber attribute API functions without writing custom code.

### Description

Adds a new `SubscriberAttributesTestingScreen` to the Compose Multiplatform tester app, accessible from the main screen. The screen is organized into five sections (User Profile, Third-party Integrations, Ad Attribution, Device, Custom Attributes), each with labeled text fields and Set/Clear buttons. Set passes the field value (blank → `null`); Clear always passes `null`. The `collectDeviceIdentifiers()` function is a single button with no input, and `setAttributes` uses a key/value pair with both buttons disabled when the key is blank.


tonidero (approved):
Nice!!!

PR #861, opened 2026-05-20
Split Swift CI to Xcode 16.4 for Xcode 26.0.1 compatibility

### LINKED ISSUE #859: RevenueCatUI in 3.0.1 references private SwiftUI symbols and SwiftUICore; iOS host link fails with "not an allowed client of SwiftUICore" + undefined SwiftUI internal symbols
- [X] I have updated Purchases SDK to the latest version
- [X] I have read
- [X] I have searched the [Community](https://community.revenuecat.com)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched

**Describe the bug**

After upgrading `purchases-kmp` from `2.10.2+17.55.1` to `3.0.1` and following the 3.0.0 migration guide (removing `PurchasesHybridCommon` / `PurchasesHybridCommonUI` from SPM and project linker), the iOS host app fails to link against the shared Kotlin/Native static framework. The Kotlin compile + `linkDebugFrameworkIosSimulatorArm64` step succeeds; the failure is at the Xcode link step for the host app.

The linker rejects an implicit reference to `SwiftUICore` (an iOS 18+ private framework with a closed `allowable-clients` list) that propagates from the precompiled `RevenueCatUI` `.o` files inside the shared framework, and then reports undefined symbols for SwiftUI internal/SPI types referenced by `RevenueCatUI` Swift sources.

Linker output (relevant excerpt):

ld: warning: object file (.../Shared.framework/Shared3379 (libicu.icudtl_dat.o)) was built for newer 'iOS-simulator' version (18.5) than being linked (16.0)
ld: warning: Could not parse or use implicit file '.../iPhoneSimulator.sdk/System/Library/Frameworks/SwiftUICore.framework/SwiftUICore.tbd':
  cannot link directly with 'SwiftUICore' because product being built is not an allowed client of it
Undefined symbols for architecture arm64:
  _symbolic ... 7SwiftUI15ModifiedContentV ... AA14_TaskModifier2V in Shared528 (CustomerCenterView.swift.o)
  _symbolic ... 7SwiftUI19_TaskValueModifier2V ... in Shared554 (PaywallView.swift.o)
  _symbolic ... AA14_TaskModifier2V in Shared590 (View+PresentPaywall.swift.o)
  _symbolic ... AA14_TaskModifier2V in Shared595 (Binding+Extensions.swift.o)
  _symbolic ... AA14_TaskModifier2V in Shared573 (VirtualCurrencyBalancesScreen.swift.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

Demangled, the missing symbols are SwiftUI underscored types treated as internal/SPI:

- `SwiftUI._TaskModifier2`
- `SwiftUI._TaskValueModifier2`
- `SwiftUI._BackgroundModifier`
- `SwiftUI._SafeAreaIgnoringLayout`
- `SwiftUI._InsettableBackgroundShapeModifier`
- `SwiftUI._PreferenceActionModifier`
- `SwiftUI._AppearanceActionModifier`
- `SwiftUI._TraitWritingModifier`
- `SwiftUI._OpacityEffect`, `_MaskEffect`, `_FrameLayout`, `_FixedSizeLayout`, `_PaddingLayout`
- `SwiftUI._EnvironmentKeyWritingModifier`
- plus many `_ConditionalContent` / `_ViewModifier_Content` references

Referenced from these `RevenueCatUI` translation units inside the shared framework:

- `CustomerCenterView.swift.o`
- `PaywallView.swift.o`
- `View+PresentPaywall.swift.o`
- `VirtualCurrencyBalancesScreen.swift.o`
- `Binding+Extensions.swift.o`

1. Environment 
    1. Platform: iOS (Kotlin Multiplatform shared framework, `isStatic = true`, consumed by a SwiftUI iOS host app via `embedAndSignAppleFrameworkForXcode`)
    2. SDK version: `purchases-kmp` 3.0.1
    3. OS version: iOS Simulator 26.0 (iPhone 17 Pro), host deployment target iOS 16.0
    4. IDE (e.g. Android Studio, Xcode, Fleet): Xcode
    5. IDE version: Xcode 26.0.1
        - [x] Device
        - [x] Emulator/simulator
    7. Environment:
        - [ ] TestFlight
        - [ ] Production
    8. How widespread is the issue. Percentage of devices affected: 100% of iOS simulator and real device builds.

2. Debug logs

Full ld diagnostic captured at the host app link step (Xcode 26.0.1, simulator arm64), reproduced verbatim above. No application-side runtime logs. Failure is at link time, the executable is never produced.

Starting from a working KMP project on `purchases-kmp` 2.10.2+17.55.1 with `PurchasesHybridCommon` + `PurchasesHybridCommonUI` wired via SPM:

  1. Bump `purchases-kmp` to `3.0.1` in the Gradle version catalog.
  2. Per the 3.0.0 migration guide, remove `PurchasesHybridCommon` and `PurchasesHybridCommonUI` from `iosApp/Package.swift` `dependencies`/`targets`, and remove the corresponding `XCRemoteSwiftPackageReference` / `XCSwiftPackageProductDependency` / `PBXBuildFile` entries from `iosApp.xcodeproj/project.pbxproj`.
  3. Build the iOS host app (Eko scheme, iOS Simulator destination, Xcode 26.0.1).

Expected:
Host app links cleanly. Per RevenueCat's guidance, no additional iOS-side SPM/Cocoapods dependency should be required after removing `PurchasesHybridCommon` because RevenueCat / RevenueCatUI are now linked into the shared Kotlin/Native framework via the new `swiftPackage(...)` build path (`upstream/purchases-ios`).

Actual:
`:shared:linkDebugFrameworkIosSimulatorArm64` succeeds. `xcodebuild` on the host app fails at link with the SwiftUICore allowable-client rejection and undefined SwiftUI internal symbols listed above.

4. Other information

The undefined symbols are all underscored SwiftUI types (`_TaskModifier2`, `_TaskValueModifier2`, `_BackgroundModifier`, etc.) — these are SwiftUI SPI / private types whose Swift ABI is not exposed for general client linking. They are not part of the public SwiftUI surface and not part of the `SwiftUI.tbd` exported symbol set that third-party products may link against.

Related (not duplicates):

- #849 — different root cause (`Unable to find module dependency: 'RevenueCat'` from `import RevenueCat` in a Swift file; the present issue does not import `RevenueCat` from Swift at all).
- #759 / #762 — cinterop-side mangled symbol (`_kniprot_swiftPMImport_..._RCPurchasesDelegate`), not SwiftUI internals.
- #828 — fixed library-side Xcode 26 cinterop build; does not address consumer link.

**Additional context**

KMP framework declaration:

```kotlin
listOf(iosArm64(), iosSimulatorArm64()).forEach { iosTarget ->
    iosTarget.binaries.framework {
        baseName = "Shared"
        isStatic = true
        linkerOpts("-framework", "AudioToolbox")
    }
}

Host iosApp/Package.swift deployment target: .iOS(.v16). No import RevenueCat or import RevenueCatUI in the Swift source on the host side. RevenueCat APIs are consumed exclusively through the Kotlin shared module from common/iOS Kotlin code (purchases-core, purchases-result, purchases-ui).

### PR DESCRIPTION
**Pull request by ajpallares**
### Checklist

### Motivation

Resolves #859. CircleCI deprecates Xcode 26.0.1 on [June 2, 2026](https://circleci.com/changelog/deprecation-of-xcode-16-3-0-and-26-0-1/); Swift artifacts compile on Xcode 16.4 instead.

### Description

- **`build-swift-artifacts`** (Xcode 16.4) → **`compileSwiftIosArtifacts`**; iOS jobs and **publish** attach workspace with `-PskipSwiftBuild=true`.
- **Fastfile** forwards `GRADLE_SKIP_SWIFT_BUILD_ARGS` so publish does not re-run Swift on Xcode 26.
- **`swiftDeveloperDir` / `SWIFT_DEVELOPER_DIR`**: local multi-Xcode override (CI uses separate executors).
- Tests cover aggregate `--dry-run` wiring, consecutive Gradle invocations, and skipSwiftBuild validation.

### CODE REVIEW

File: build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/task/SwiftBuildTask.kt [hunk 1]
```diff
@@ -99,6 +99,14 @@ abstract class SwiftBuildTask @Inject constructor(
     @get:Optional
     abstract val swiftSettingsArgs: ListProperty<String>
 
+    /**
+     * Optional Xcode developer directory passed to `DEVELOPER_DIR` when compiling Swift.
+     * Local-dev escape hatch when multiple Xcodes are installed; CI uses separate executors.
+     */
+    @get:Input
+    @get:Optional
+    abstract val developerDir: Property<String>
```
JayShortway: Is `DEVELOPER_DIR` a standard env var set by Xcode?

.circleci/config.yml
```diff
@@ -125,6 +140,8 @@ executors:
       # Avoid waiting for Homebrew to auto update existing packages, as everything we care about is
       # freshly installed.
       HOMEBREW_NO_AUTO_UPDATE: 1
+      # Precompiled Swift static libraries are built on Xcode 16.4 and consumed by Kotlin/Native.
+      GRADLE_SKIP_SWIFT_BUILD_ARGS: "-PskipSwiftBuild=true -PswiftConfiguration=release"
```
JayShortway, line 144: I'm not sure this is a good idea. We're now having to pass this to every Gradle invocation that could possibly compile the iOS target.

I think it would be better to configure this behavior in `local.properties`. Maybe we have a key that points to the directory containing the precompiled libraries (`revenuecat.precompiledSwiftLibrariesDirectory` or something, or maybe we need 2), and use the existence of that key to determine whether to skip the Swift build or not.

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/ConfigureSwiftDependencies.kt
```diff
@@ -252,9 +258,23 @@ private fun Project.registerSwiftBuildTask(
     targetSourceDir: File,
     transitiveDepSourceDirs: List<File>,
     moduleDependencies: List<GlobalSwiftPackageRegistry.RegisteredSwiftTarget>,
-): TaskProvider<SwiftBuildTask> {
+    swiftOutputDir: Provider<Directory>,
+): TaskProvider<out Task> {
     val taskName = compileSwiftTaskName(dependency.target, kotlinTarget.konanTarget)
 
+    if (shouldSkipSwiftBuild()) {
```
JayShortway: I don't think the `shouldSkipSwiftBuild()` check should be _inside_ the `registerSwiftBuildTask()` function. That should've been checked beforehand.

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/SwiftBuildProperties.kt
```diff
@@ -0,0 +1,44 @@
...
+ * Optional override for the Xcode used to compile Swift static libraries.
+ *
+ * Intended for local development when multiple Xcode versions are installed (e.g.
+ * `xcode-select` points at 26.x but Swift artifacts should be built with 16.4).
+ * CI uses dedicated executors instead (`xcode16` vs `xcode26`).
+ */
+internal fun Project.getSwiftDeveloperDir(): String? =
+    findProperty("swiftDeveloperDir")?.toString()?.takeIf { it.isNotBlank() }
+        ?: providers.environmentVariable("SWIFT_DEVELOPER_DIR").orNull?.takeIf { it.isNotBlank() }
+
```
JayShortway: How does this work? I'm not sure if we should care about this for local development (for now), to avoid some complexity.

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/task/ValidatePrecompiledSwiftArtifactsTask.kt
```diff
@@ -0,0 +1,55 @@
...
+ * or machine.
+ *
+ * This task has no outputs; `outputs.upToDateWhen { true }` keeps Gradle from re-running it
+ * once the first validation in a build has succeeded.
+ */
+abstract class ValidatePrecompiledSwiftArtifactsTask : DefaultTask() {
+
+    init {
+        outputs.upToDateWhen { true }
+    }
```
JayShortway, line 23: This feels a bit like a hack. But also, what are we validating exactly? Wouldn't the rest of the build fail anyway if the precompiled artifacts are missing?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/SwiftBuildProperties.kt
```diff
@@ -0,0 +1,44 @@
...
+internal fun Project.ensureCompileSwiftIosArtifactsAggregateTask() {
```
JayShortway:
```suggestion
internal fun Project.ensureAggregateCompileSwiftIosArtifactsTaskRegistered() {
```

.circleci/config.yml
JayShortway, line 265: Should we run this one on 26.0 (while we can), instead of 26.4?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/task/SwiftBuildTask.kt
ajpallares: Yes. `DEVELOPER_DIR` is Apple’s standard env var for “which Xcode.app to use.” `xcodebuild`, `xcrun`, and `swift` all respect it. `xcode-select` effectively sets the default. See https://developer.apple.com/documentation/xcode/configuring-command-line-tools-settings?changes=_8,_8#Select-the-default-Xcode-app-temporarily

.circleci/config.yml
ajpallares: Fair point. We shouldn't have to append skip flags on every Gradle call.

Regarding the `local.properties` approach for this, since it's `gitignore`d, we'd need new build-logic to read it, CI would still have to generate it each job, and locally, the "key exists --> skip build" logic could silently reuse stale artifacts in `build/`.

Would `ORG_GRADLE_PROJECT_skipSwiftBuild=true` (and `swiftConfiguration=release`) on the `xcode26` executor work instead? Same explicit behavior, but it would be centralized. No `${GRADLE_SKIP_SWIFT_BUILD_ARGS}` on every `./gradlew` or Fastlane call.

WDYT @JayShortway? I honestly don't have a strong preference

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/task/SwiftBuildTask.kt
ajpallares: Anyway, this was removed in https://github.com/RevenueCat/purchases-kmp/pull/861/commits/780a3c3c4c7953551a2b711798e061b20c99cec8 to address [this other comment](https://github.com/RevenueCat/purchases-kmp/pull/861#discussion_r3280256484)

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/ConfigureSwiftDependencies.kt
ajpallares: Addressed in https://github.com/RevenueCat/purchases-kmp/pull/861/commits/780a3c3c4c7953551a2b711798e061b20c99cec8

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/SwiftBuildProperties.kt
ajpallares: Right. Removed in https://github.com/RevenueCat/purchases-kmp/pull/861/commits/780a3c3c4c7953551a2b711798e061b20c99cec8

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/task/ValidatePrecompiledSwiftArtifactsTask.kt
ajpallares: You're right that the build would fail anyway if the precompiled artifacts are missing. The difference is when and how. Without this task, we'd get a vague cinterop/link error later (mostly when running locally with`-PskipSwiftBuild=true` without building artifacts first). The `upToDateWhen { true }` part is admittedly hacky...

With this, we fail immediately with a clear "missing precompiled Swift library at …" message instead of getting a later with a more confusing Clang/module errors. That's the main purpose of this. 

I guess we are OK dropping the task and accepting the worse error in that scenario, since:
1. Locally we wouldn't normaly use `PskipSwiftBuild=true`
2. On CI, it isn't likely to happen that we don't have the files present

WDYT?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/SwiftBuildProperties.kt
ajpallares: Done in https://github.com/RevenueCat/purchases-kmp/pull/861/commits/780a3c3c4c7953551a2b711798e061b20c99cec8

.circleci/config.yml
ajpallares: What would be the benefit? I see that 26.0 would best simulate the reporter's environment at link time, but we're not running the sample app across Xcode versions anyway. If we have to pick one, I'd rather stay on a more recent Xcode (especially since 26.0.1 is being deprecated on CircleCI soon)

ajpallares:
Closing in favor of #872

PR #862, opened 2026-05-20
[AUTOMATIC] Release/3.0.2

RCGitBot:
**This is an automatic release.**

### 📦 Dependency Updates
* [RENOVATE] Update dependency gradle to v9.5.1 (#857) via RevenueCat Git Bot (@RCGitBot)
* [RENOVATE] Update dependency upstream/purchases-ios to v5.73.0 (#858) via RevenueCat Git Bot (@RCGitBot)

PR #863, opened 2026-05-21
Fix VERSIONS.md not updating on automatic releases

ajpallares:
### Checklist

### Motivation

`VERSIONS.md` is missing the 3.0.1 row. The `automatic-bump` CircleCI job does not initialize submodules, so `get_ios_version` crashed with `Errno::ENOENT` reading `upstream/purchases-ios/.version`. The crash happens after the release PR has been pushed, leaving the lane unable to emit the follow-up `Updates VERSIONS.md` commit.

### Description

`get_ios_version` now reads `.version` from the local submodule checkout when available, and falls back to fetching it from the GitHub Contents API pinned to the submodule's commit SHA. Mirrors the `get_billing_client_version` pattern, but avoids an extra API call when the submodule is already initialized.

Also backfills the 3.0.1 row in `VERSIONS.md`. The 3.0.2 row from #862 will need to be backfilled separately if it merges before this lands.


JayShortway (approved):
Nice!!

PR #873, opened 2026-05-26
Add Maven consumer link checks to release CI

ajpallares:
### Checklist

  hybrids

### Motivation

Follow-up to #872. Guards against #859 by validating Maven publish output and consumer linking before tagging a release.

### Description

- Release-only publish + Android/iOS consumer link jobs; `installationTestApp` uses Maven Local in those jobs, project deps otherwise
- Both consumer jobs gate the release hold
- Verified: [successful CI run](https://app.circleci.com/pipelines/gh/RevenueCat/purchases-kmp/3226/workflows/d8c1800c-eea0-40e3-944b-32cb1227c5b2)

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -12,8 +12,14 @@ pluginManagement {
 
 dependencyResolutionManagement {
     repositories {
+        val gradleProperties = java.util.Properties()
+        val gradlePropertiesFile = File(settingsDir, "gradle.properties")
+        if (gradlePropertiesFile.exists()) {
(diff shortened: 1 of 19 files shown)


settings.gradle.kts
```diff
@@ -12,11 +12,25 @@ pluginManagement {
 
 dependencyResolutionManagement {
     repositories {
+        if (usePublishedMavenLocalArtifacts()) {
+            mavenLocal()
+        }
```
ajpallares, line 17: I'm not sure this is the best approach, since this could bite us locally. Do you think I should scope this only for the installation test app?

PR #896, opened 2026-06-11
Replace negative lookahead in CircleCI `when: matches:` condition

ajpallares:
### Motivation

CircleCI is [updating the regex engine used for config compilation](https://circleci.com/changelog/breaking-change-regex-engine-update-for-config-compilation/) on June 26, 2026. After that date, negative lookaheads in `when: matches: pattern:` conditions will fail to compile, breaking this pipeline.

### Description

Rewrites the `release-tags-workflow` condition as an equivalent `matches` + `not: matches` combination. The `release-tags-job` tag filter regex is unchanged: job filters are not affected by this change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,12 +1,16 @@
 aliases:
-  # Uses the same regex as release-tags-job
+  # Matches the same tags as release-tags-job, but without lookaheads,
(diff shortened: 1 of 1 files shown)


tonidero (approved):
Thank you! 🙇

PR #901, opened 2026-06-15
Add presentedOfferingContext support to custom paywall impression events

rickvdl:
Based on PHC PR https://github.com/RevenueCat/purchases-hybrid-common/pull/1665.

## API

We will now send the `presentedOfferingContext` from the passed `Offering` along with the event. This PR adds the new API variant that takes an `Offering` instead of a plain string `offeringId`. The latter is now deprecated in favor of this new API variant. We will derive the `presentedOfferingContext` from this `Offering` and send it along with the request.

```kotlin
Purchases.sharedInstance.trackCustomPaywallImpression(
    CustomPaywallImpressionParams(paywallId = "", offering = offering),
)
```

### DIFF
--- a/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
+++ b/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
@@ -4,10 +4,97 @@ package com.revenuecat.purchases.kmp.models
  * Parameters for tracking a custom paywall impression event.
  *
  * @param paywallId An optional identifier for the custom paywall being shown.
- * @param offeringId An optional identifier for the offering associated with the custom paywall.
- * If not provided, the SDK will use the current offering identifier from the cache.
+ * @param offeringId [Deprecated] Use [offering] instead. Optional identifier for the offering associated
+ * with the custom paywall.
+ * @property offering The [Offering] associated with the custom paywall, if provided.
  */
-public class CustomPaywallImpressionParams(
+public class CustomPaywallImpressionParams private constructor(
     public val paywallId: String? = null,
+    @Deprecated(
+        message = "Pass an Offering object instead. Using an offering identifier string prevents " +
+            "the SDK from deriving placement and targeting context automatically.",
+    )
     public val offeringId: String? = null,
-)
+    public val offering: Offering?,
+) {
+    /**
+     * Creates parameters with no offering information. No offering data is sent and the native SDK
+     * chooses the current offering.
+     */
+    public constructor() : this(
+        paywallId = null,
+        offeringId = null,
+        offering = null,
+    )
+
+    /**
+     * Creates parameters with only a paywall identifier. No offering data is sent and the native SDK
+     * chooses the current offering.
+     *
+     * @param paywallId An optional identifier for the custom paywall being shown.
+     */
+    public constructor(
+        paywallId: String?,
+    ) : this(
+        paywallId = paywallId,
+        offeringId = null,
+        offering = null,
+    )
+
+    /**
+     * Creates parameters with an offering identifier string.
+     *
+     * @param paywallId An optional identifier for the custom paywall being shown.
+     * @param offeringId [Deprecated] Use [offering] instead. Optional identifier for the offering associated
+     * with the custom paywall.
+     * @deprecated Pass an [Offering] object instead. Using an offering identifier string prevents
+     * the SDK from deriving placement and targeting context automatically.
+     */
+    @Deprecated(
+        message = "Pass an Offering object instead. Using an offering identifier string prevents " +
+            "the SDK from deriving placement and targeting context automatically.",
+    )
+    public constructor(
+        paywallId: String? = null,
+        offeringId: String?,
+    ) : this(
+        paywallId = paywallId,
+        offeringId = offeringId,
+        offering = null,
+    )
+
+    /**
+     * Creates parameters from the offering the custom paywall was obtained from.
+     *
+     * Use this constructor when presenting a paywall for an offering that is not the current
+     * offering, for example a placement-resolved offering. The SDK will derive both the offering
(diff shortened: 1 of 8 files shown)


models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
```diff
@@ -4,10 +4,97 @@ package com.revenuecat.purchases.kmp.models
...
+        message = "Pass only an Offering object instead. The offeringId is ignored when Offering is provided.",
+        replaceWith = ReplaceWith("CustomPaywallImpressionParams(offering, paywallId)"),
+    )
+    public constructor(
+        paywallId: String? = null,
+        offering: Offering,
+        @Suppress("UNUSED_PARAMETER")
+        offeringId: String?,
+    ) : this(paywallId, offering.identifier, offering)
+}
```
ajpallares: If `offering` is a new parameter we're adding, why do we need to even declare this constructor if it wasn't available before?

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
```diff
@@ -4,10 +4,97 @@ package com.revenuecat.purchases.kmp.models
...
+     * @param paywallId An optional identifier for the custom paywall being shown.
+     * @param offeringId [Deprecated] Use [offering] instead. Optional identifier for the offering associated
+     * with the custom paywall.
+     * @deprecated Pass an [Offering] object instead. Using an offering identifier string prevents
+     * the SDK from deriving placement and targeting context automatically.
+     */
+    @Deprecated(
+        message = "Pass an Offering object instead. Using an offering identifier string prevents " +
+            "the SDK from deriving placement and targeting context automatically.",
+    )
```
ajpallares, line 57: Shouldn't we add this here?

```suggestion
    @Deprecated(
        message = "Pass an Offering object instead. Using an offering identifier string prevents " +
            "the SDK from deriving placement and targeting context automatically.",
        replaceWith = ReplaceWith("CustomPaywallImpressionParams(offering, paywallId)"),
    )
```

ajpallares (commented):
I think it looks good! But I have one question about the new APIs, specially about one specific init that we may not need to declare

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
```diff
@@ -4,10 +4,97 @@ package com.revenuecat.purchases.kmp.models
...
+     * Use this constructor when presenting a paywall for an offering that is not the current
+     * offering, for example a placement-resolved offering. The SDK will derive both the offering
+     * identifier and the presented offering context from the provided offering.
+     *
+     * @param paywallId An optional identifier for the custom paywall being shown.
+     * @param offering The [Offering] associated with the custom paywall.
+     */
+    public constructor(
+        offering: Offering,
+        paywallId: String? = null,
```
tonidero: Very nitpicky but I wonder if we should kep the paywallId as the first parameter to be consistent with the other constructors...

tonidero (commented):
Agree with the existing comments and added one as well. Other than that, looking great!

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
rickvdl: Good suggestion, added, thanks!

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
rickvdl: Good point, my Codex review flagged this as "more kotlin like", I wasn't entirely sure so went with this and in the API tester / examples I kept it consistent. But I think I agree now that it makes sense to just keep it consistent. Updated

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
rickvdl: Valid point! I think I added it for completeness in the API tests, but this doesn't really make sense 😅 Thanks for flagging

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
```diff
@@ -4,10 +4,78 @@ package com.revenuecat.purchases.kmp.models
...
-public class CustomPaywallImpressionParams(
+public class CustomPaywallImpressionParams private constructor(
     public val paywallId: String? = null,
+    @Deprecated(
+        message = "Pass an Offering object instead. Using an offering identifier string prevents " +
+            "the SDK from deriving placement and targeting context automatically.",
+    )
     public val offeringId: String? = null,
-)
+    public val offering: Offering?,
```
tonidero, line 18: It's a bit different than in Unity, for example, where we don't expose the `offering` object but expose the `PresentedOfferingContext` directly... TBH, I kinda think this is better... but not a big deal :)

tonidero (approved):
Seems we need to update the API files. But other than that, I think this makes sense!

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/CustomPaywallImpressionParams.kt
rickvdl: Hm you're right indeed, and the others should expose Offering as well. Will look into Unity to see if we can solve the inconsistency there.

PR #904, opened 2026-06-17
Generate KMP native error-code mappings

AlvaroBrey:
Generates the native -> KMP error-code mappings (`errors.android.kt`, `errors.ios.kt`) from the purchases-error-codes SSOT. Closes SDK-4377.


ajpallares (approved):
🙌

PR #922, opened 2026-07-03
Add preferred UI locale override support

AlvaroBrey:
`purchases-kmp` is missing the preferred UI locale override that the other SDKs offer (config option + runtime setter). Found while auditing hybrids for [SDK-4390](https://linear.app/revenuecat/issue/SDK-4390/support-preferred-ui-locale-override-in-unity-sdk).

Implements both config-time and runtime accessors, mirroring the other SDKs.

Depends on https://github.com/RevenueCat/purchases-ios/pull/7121 for objC bindings for the locale accessors.

Closes SDK-4394.

---

> [!NOTE]
> **Low Risk**
> Additive public API and thin delegates to existing native SDK behavior; no changes to purchase, auth, or entitlement flows.
> 
> **Overview**
> Adds **preferred UI locale override** to `purchases-kmp`, matching other RevenueCat SDKs so Paywalls and related UI can use a locale like `de_DE` instead of the device default.
> 
> **Configure-time:** `PurchasesConfiguration.Builder.preferredUILocaleOverride` (nullable, default `null`) is passed through on Android and iOS during `configure`.
> 
> **Runtime:** `Purchases.overridePreferredUILocale(locale)` sets or clears the override after configure; documented to affect RevenueCat UI components such as Paywalls.
> 
> Public ABI files (`core.api`, `core.klib.api`), API tester samples, and iOS configuration tests cover the new surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3609690ffc2ae0345141b172895a9af92d32ab6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
```diff
@@ -131,6 +131,7 @@ public actual class Purchases private constructor(private val androidPurchases:
                         .pendingTransactionsForPrepaidPlansEnabled(
                             pendingTransactionsForPrepaidPlansEnabled ?: false
                         )
+                        .apply { preferredUILocaleOverride?.let { preferredUILocaleOverride(it) } }
```
ajpallares: Hmm doesn't the `preferredUILocaleOverride` method already receives an optional value?
```suggestion
                        .preferredUILocaleOverride(preferredUILocaleOverride)
```

In that case, we should probably pass it as it. We Shouldn't assume here that the SDK has no specific different behavior if this parameter is pass to `null`.

ajpallares (commented):
Looking great! Firstly, one thing I think should be changed

core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt
AlvaroBrey: Thanks! Yes, this was an oversight on my side, in fact it could be called a bug since it would not let KMP users _remove_ an override. Good eye!

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
```diff
@@ -131,6 +132,7 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
                     .withDangerousSettings(dangerousSettings.toIosDangerousSettings())
                     .withShowStoreMessagesAutomatically(showInAppMessagesAutomatically)
                     .withEntitlementVerificationMode(verificationMode.toIosEntitlementVerificationMode())
+                    .apply { preferredUILocaleOverride?.let { withPreferredUILocaleOverride(it) } }
```
ajpallares: Same comment here

AlvaroBrey: Hmmm so prompted by this I was looking into what the other hybrids are doing with nulls, and found that PHC [actually skips nulls](https://github.com/RevenueCat/purchases-hybrid-common/pull/1267#discussion_r2322492167) 🙃 and has done so since this feature exists

So it seems like we've a choice to make

ajpallares: Oh great catch! Totally missed it. IMO, this should be handled by the natives and the hybrid layers (including PHC) should not deal with it. 

This way, we:
1. Minimize the logic in the hybrids, and
2. We make sure the null cases are handled by the natives. E.g what if call configure twice and pass null in the second configure call? (Configure shouldn't be called twice, but theoretically I mean)

tonidero: Agreed... I think that's a bug right now yeah. Great catch!!

tonidero (commented):
Only that remaining comment and I think this should be good! 🙏

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
AlvaroBrey: From my understanding, in PHC a `null` at config time is interpreted as `not set/default`.

This makes sense for other fields that are not nullable per se, but kinda doesn't with this `String?`.  We have no way to distinguish "explicitly set to null" vs "not set, use the default".
It just happens to make sense because the default is `null`, but if it was a nullable with a non-null default we would be in a strange place. What we really should've done is set a explicit non-nullable string with a default value like `"DEFAULT"`  or so, and keep `null` to mean "not set` on hybrids only.

Either way this is an interesting topic for API design going forward, but what should we do with KMP?
a) Leave the skip: consistent with PHC but not needed, since the default is null anyway in the builder
b) Remove the skip: nconsistent with PHC, but lets natives decide what to do with hybrids

This is more of a philosophical question since if we were to make the default something other than `null` it would be breaking behaviour anyway. But I think I'll remove the skip here

AlvaroBrey: Done!

tonidero: Yeah, basically the difference between `null` and `undefined` 😂. I think that's a very valid discussion, since removing the skip basically forces us to duplicate the defaults in the hybrids... 
I would say for nullable properties like this, the common default is likely `null` itself. Otherwise, we would represent the setting with an enum-like object I would say, so I think removing the skip for nullable properties is okish IMO. But yeah, good discussion to have!

tonidero (approved):
Nice!!

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt
ajpallares: Agreed with Toni here! It is a great discussion indeed, and something I hadn't thought of tbh...
Let's do that with the `null`s, since the common default is indeed `null`.

ajpallares (approved):

PR #943, opened 2026-07-23
Add internal support for enabling workflows

vegaro:
## What changed

Adds an experimental `useWorkflows` field to KMP `DangerousSettings`, defaulting to `false`, and maps it to the native SDK configuration:

- Android uses the internal `DangerousSettings.forWorkflows(autoSyncPurchases)` API.
- iOS uses a small Swift factory to reach the `@_spi(Internal)` workflow initializer, which is unavailable through the generated bindings.

The API-tester fixture covers the new experimental field.

## Verification
- [x] `:core:compileKotlinIosArm64`
- [x] `:core:compileDebugKotlinAndroid`
- [x] `:apiTester:compileKotlinIosArm64`

This mirrors purchases-unity#996 and react-native-purchases#1847. Defaults remain unchanged when the option is omitted.


vegaro:
This will probably be used only for tests

PR #912, opened 2026-06-22
Use mise for JDK management

AlvaroBrey:
Enables `mise` to manage the JDK for this project, aligning with other repos like `https://github.com/RevenueCat/purchases-ios` that already use `mise`. A step toward a single env setup tool.

kmp didn't actually have a  `.sdkmanrc`, so this is purely additive

---

> [!NOTE]
> **Low Risk**
> Documentation and local dev tooling only; no SDK, CI, or runtime code paths are modified.
> 
> **Overview**
> **Developer environment setup** now uses **mise** instead of sdkman for the local JDK, matching other RevenueCat repos.
> 
> New **`mise.toml`** pins **`liberica-17.0.6`** with lockfile enabled, and **`mise.lock`** records the resolved **`liberica-17.0.6+10`** install. **`AGENTS.md`** step 1 tells contributors to install mise and run **`mise install`** in the project root.
> 
> No application or Gradle build logic changes—only onboarding and toolchain pinning for local development.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bf71debc15448fd1c1ca5d473b30e2086aa46ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #947, opened 2026-07-27
Wire mise into CI for JDK and Ruby

AlvaroBrey:
Wires the JDK/Ruby pins already declared in `mise.toml` (added in #912 for local dev only) into CI, so CI actually uses them instead of whatever the executor image happens to ship.

- Adds `install-mise-tools` (`tools: java`) to all 13 Gradle-invoking jobs. Previously there was zero explicit JDK pin in CI: both the Linux (`android202409`) and macOS (`xcode16`) executors just used the image's ambient JDK.
- Adds `ruby = "3.3.0"` to `mise.toml` and wires `tools: java,ruby` into the 5 jobs that also call `install-gem-mac-dependencies` on macOS. That command never provisioned ruby itself, so macOS ruby was previously fully unpinned (no rbenv, no cimg image). Matches the Linux `ruby3` executor's existing `cimg/ruby:3.3.0`. `rb-readline` is already a direct `Gemfile` dependency, so no new readline-crash risk (fastlane#21794) from this.
- Regenerates `mise.lock` for all 4 platforms (linux/macos × arm64/x64) — it previously had version-only entries with no platform checksums.

---

> [!NOTE]
> **Low Risk**
> CI-only toolchain pinning with no application or SDK runtime changes; risk is mainly CI flakiness if mise installs fail.
> 
> **Overview**
> CI now provisions **Java** and **Ruby** from `mise.toml` / `mise.lock` instead of relying on CircleCI executor images.
> 
> **CircleCI** adds `revenuecat/install-mise-tools` before Gradle (and gem steps on macOS): `tools: java` on all Gradle jobs, and `tools: java,ruby` on macOS jobs that use `install-gem-mac-dependencies` plus `publish` and `update-error-codes` API regeneration.
> 
> **mise.toml** pins **Ruby 3.3.0** (with `compile = false`) alongside the existing Liberica JDK pin. **mise.lock** is expanded with per-platform checksums and download URLs for Java and Ruby on linux/macos arm64 and x64.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 195f88e24a77f852f0268087cc4b0b88157b81de. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,6 +163,8 @@ jobs:
     steps:
       - checkout
       - checkout-submodule
+      - revenuecat/install-mise-tools:
+          tools: java
       - restore-gradle-user-home-directory-from-cache
       - run:
           name: Run Detekt on commonMain
@@ -181,6 +183,8 @@ jobs:
       - install-android-sdk-on-macos
       - checkout
       - checkout-submodule
+      - revenuecat/install-mise-tools:
+          tools: java
       - restore-gradle-user-home-directory-from-cache
       - restore-kotlin-native-compiler-from-cache
       - attach-workspace-at-working-directory
@@ -194,6 +198,8 @@ jobs:
     executor: android202409
     steps:
       - checkout
+      - revenuecat/install-mise-tools:
+          tools: java
       - restore-gradle-user-home-directory-from-cache
       - attach-workspace-at-working-directory
       - run:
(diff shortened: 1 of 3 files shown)


tonidero (approved):
🙌

PR #962, opened 2026-08-05
build-logic groundwork for watchOS targets

AlvaroBrey:
Part of #126. No module uses any of this yet, so existing targets build identically.

### Motivation

Groundwork in build-logic for adding watchOS targets in the follow-up PRs.

### Description

- Fixes the Swift target triple for `watchosArm64`: Kotlin/Native maps it to `arm64_32-apple-watchos` (32-bit `NSInteger`), not `arm64`. `watchosDeviceArm64` is the arm64 one.
- Adds watchOS linker opts to the generated cinterop def file (Swift runtime search paths).
- New `watchosTargets()` helper for build scripts.
- Generalizes the Apple opt-in wiring to all Apple source-set prefixes and adds `kotlinx.cinterop.UnsafeNumber` (needed once variable-width `NSInteger` bindings are commonized across 32/64-bit targets).

---

> [!NOTE]
> **Low Risk**
> Build-logic only; no library modules enable watchOS yet. The main behavioral change is extra cinterop opt-ins on Apple source sets.
> 
> **Overview**
> Adds **build-logic groundwork** for watchOS without enabling it in any module yet.
> 
> Introduces `watchosTargets()` (`watchosArm64`, `watchosDeviceArm64`, `watchosSimulatorArm64`) for modules that opt in later.
> 
> Swift/cinterop wiring drops Intel `watchosX64`, uses the correct `arm64_32` triple for `watchosArm64` vs `arm64` for `watchosDeviceArm64`, and adds watchOS Swift runtime `linkerOpts` in generated `.def` files.
> 
> Apple language settings now apply `ExperimentalForeignApi` (and new `UnsafeNumber`) to all Apple source-set prefixes, not just iOS, so commonized 32/64-bit `NSInteger` bindings can compile.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 421085d34376de61497ca36052ca80d045cbfb30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/AppleTargets.kt
+++ b/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/AppleTargets.kt
@@ -1,0 +1,14 @@
+package com.revenuecat.purchases.kmp.buildlogic
+
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+
+/**
+ * Declares all watchOS compilation targets. The `revenuecat-library` convention plugin declares
+ * Android and iOS targets for every module; modules that also support watchOS call this from
+ * their `kotlin {}` block.
+ */
+fun KotlinMultiplatformExtension.watchosTargets() {
+    watchosArm64()
+    watchosDeviceArm64()
+    watchosSimulatorArm64()
+}
--- a/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
+++ b/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
@@ -35,8 +35,12 @@ internal fun Project.configureKotlin() {
         }
         sourceSets.all {
             languageSettings.apply {
-                if (name.lowercase().startsWith("ios")) {
+                val appleSourceSetPrefixes = listOf("apple", "ios", "watchos", "tvos", "macos")
+                if (appleSourceSetPrefixes.any { name.lowercase().startsWith(it) }) {
                     optIn("kotlinx.cinterop.ExperimentalForeignApi")
+                    // NSInteger and friends commonize to variable-width types (32 bits on
+                    // watchosArm64), which require this opt-in in shared Apple code.
+                    optIn("kotlinx.cinterop.UnsafeNumber")
                 }
             }
         }
--- a/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
+++ b/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
@@ -38,8 +38,9 @@ internal fun Project.configureKotlin() {
                 val appleSourceSetPrefixes = listOf("apple", "ios", "watchos", "tvos", "macos")
                 if (appleSourceSetPrefixes.any { name.lowercase().startsWith(it) }) {
                     optIn("kotlinx.cinterop.ExperimentalForeignApi")
-                    // NSInteger and friends commonize to variable-width types (32 bits on
-                    // watchosArm64), which require this opt-in in shared Apple code.
+                    // NSInteger and friends commonize to types of different widths on different
+                    // Apple targets (e.g. 64 bits on iOS, 32 bits on watchosArm64), which requires
+                    // this opt-in in shared Apple code.
                     optIn("kotlinx.cinterop.UnsafeNumber")
                 }
             }
--- a/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/ConfigureSwiftDependencies.kt
+++ b/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/ConfigureSwiftDependencies.kt
@@ -219,7 +219,6 @@ private fun KonanTarget.isAppleTarget(): Boolean =
         // watchOS
         KonanTarget.WATCHOS_ARM64,
         KonanTarget.WATCHOS_SIMULATOR_ARM64,
-        KonanTarget.WATCHOS_X64,
         KonanTarget.WATCHOS_DEVICE_ARM64,
     )
 
@@ -310,7 +309,6 @@ private fun getTaskSuffix(konanTarget: KonanTarget): String = when (konanTarget)
     // watchOS
     KonanTarget.WATCHOS_ARM64 -> "WatchosArm64"
     KonanTarget.WATCHOS_SIMULATOR_ARM64 -> "WatchosSimulatorArm64"
-    KonanTarget.WATCHOS_X64 -> "WatchosX64"
     KonanTarget.WATCHOS_DEVICE_ARM64 -> "WatchosDeviceArm64"
     else -> error("Unexpected target: $konanTarget")
 }
@@ -376,7 +374,7 @@ private fun KonanTarget.getSdkName(): String = when (this) {
     KonanTarget.TVOS_X64, KonanTarget.TVOS_SIMULATOR_ARM64 -> "appletvsimulator"
     KonanTarget.TVOS_ARM64 -> "appletvos"
     // watchOS
-    KonanTarget.WATCHOS_X64, KonanTarget.WATCHOS_SIMULATOR_ARM64 -> "watchsimulator"
+    KonanTarget.WATCHOS_SIMULATOR_ARM64 -> "watchsimulator"
     KonanTarget.WATCHOS_ARM64, KonanTarget.WATCHOS_DEVICE_ARM64 -> "watchos"
     else -> error("Unsupported target: ${this}")
 }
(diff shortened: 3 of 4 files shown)


vegaro (approved):

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
```diff
@@ -35,8 +35,12 @@ internal fun Project.configureKotlin() {
         }
         sourceSets.all {
             languageSettings.apply {
-                if (name.lowercase().startsWith("ios")) {
+                val appleSourceSetPrefixes = listOf("apple", "ios", "watchos", "tvos", "macos")
+                if (appleSourceSetPrefixes.any { name.lowercase().startsWith(it) }) {
                     optIn("kotlinx.cinterop.ExperimentalForeignApi")
+                    // NSInteger and friends commonize to variable-width types (32 bits on
+                    // watchosArm64), which require this opt-in in shared Apple code.
```
JayShortway: Had to read this a few times. Maybe something along the lines of:
> NSInteger and friends commonize to types of different widths on different Apple targets (e.g. 64 bits on X, 32 bits on watchosArm64), which requires this opt-in in shared Apple code.

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
```diff
@@ -35,8 +35,12 @@ internal fun Project.configureKotlin() {
...
+                    optIn("kotlinx.cinterop.UnsafeNumber")
```
JayShortway, line 44: Do app developers need to care about this? As in, do they need to pass the same opt-in? Is the answer to the first question dependent on having a watchOS target?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
JayShortway: Does this mean that we can't reliably convert an `NSInteger` into a Kotlin type (`Int` or `Long`) in e.g. `appleMain` without checking the platform we're actually running on?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/ConfigureSwiftDependencies.kt
```diff
@@ -394,8 +392,10 @@ private fun KonanTarget.getTriple(): String = when (this) {
     KonanTarget.TVOS_SIMULATOR_ARM64 -> "arm64-apple-tvos-simulator"
     KonanTarget.TVOS_ARM64 -> "arm64-apple-tvos"
     // watchOS
-    KonanTarget.WATCHOS_X64 -> "x86_64-apple-watchos-simulator"
```
JayShortway, line 397: Is `WATCHOS_X64` deprecated? Or is there another reason this was removed?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/AppleTargets.kt
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Declares all watchOS compilation targets. The `revenuecat-library` convention plugin declares
+ * Android and iOS targets for every module; modules that also support watchOS call this from
+ * their `kotlin {}` block.
+ */
+fun KotlinMultiplatformExtension.watchosTargets() {
+    watchosArm64()
+    watchosDeviceArm64()
+    watchosSimulatorArm64()
+}
```
JayShortway, line 14: Which modules that support iOS won't support watchOS?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/ConfigureSwiftDependencies.kt
```diff
@@ -394,8 +392,10 @@ private fun KonanTarget.getTriple(): String = when (this) {
...
     KonanTarget.WATCHOS_SIMULATOR_ARM64 -> "arm64-apple-watchos-simulator"
-    KonanTarget.WATCHOS_ARM64, KonanTarget.WATCHOS_DEVICE_ARM64 -> "arm64-apple-watchos"
+    // Note: Kotlin/Native's watchosArm64 is arm64_32 (see targetTriple.watchos_arm64 in
+    // konan.properties), while watchosDeviceArm64 is arm64.
```
JayShortway: Where is `konan.properties`? It's not in this repo right?

JayShortway (commented):
Looks good! Just have a few questions. Might be some dumb ones, but it's been a while 😅

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/ConfigureSwiftDependencies.kt
AlvaroBrey: Right, deprecated since Kotlin 2.3

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/AppleTargets.kt
AlvaroBrey: UI modules basically (no compose deps available for watchOS)

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
AlvaroBrey: addressed (just copied your text)

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
AlvaroBrey: Nope, this only affects our shared code. Consumers compile against the platform-specific type which is properly resolved, and by the time anything reaches the public API it's a plain fixed-width Long/Int

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
AlvaroBrey: Hmm sort of.

* when `appleMain` compiles for iOS -> `NSInteger = Long`
* when it compiles for arm64_32 -> `NSInteger = Int`

Most of the problems this could cause are resolved at compile time though, not runtime, so we are mostly safe.
For example [this line](https://github.com/RevenueCat/purchases-kmp/pull/964/changes#diff-30185c48f0b657adb2563d3a4f3a5f0270c3f8b5b50f6fbc03c8fb8ecd797f64R20) fails compilation for watchOS arm64_32 if you remove the `toLong`, since `NSInteger` resolves to `Int` and can't be assigned to `Long`.
 
However this is all kind of a mess tbh. I blame 🍎 . 

And btw, to make matters even more fun, there is no arm64_32 simulator 👍 The only way to do real verification is with a physical device

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
AlvaroBrey: https://github.com/RevenueCat/purchases-kmp/pull/962#discussion_r3795756780 for related explanation

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/swift/ConfigureSwiftDependencies.kt
AlvaroBrey: Yeah, it's in the konan data folder.  I'm just gonna remove the comment because the literals tell the whole story

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
JayShortway: Ahh okay that's a mess indeed. I guess the advice then is to always call `toLong()` on any `NSInteger`?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/AppleTargets.kt
JayShortway: Got it!

JayShortway (approved):
Looks good! Thanks for answering my questions!

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/kmp/buildlogic/convention/configureKotlin.kt
AlvaroBrey: If you already have a `NSInteger` and you need a `Long`, then yes.  In some cases like [this one](https://github.com/RevenueCat/purchases-kmp/pull/964#discussion_r3805053337) when what you have is a `NsDecimalNumber` you should use `longLongValue` instead, because `longValue.toLong()` would trim it to 32 bits and then expand it to 64 again. 

why are you like this apple

PR #964, opened 2026-08-05
Add watchOS support

### LINKED ISSUE #126: WatchOS target

AlvaroBrey:
Resolves #126 for watchOS. tvOS and macOS are left for follow-ups.

### Checklist

  hybrids

### Description

Adds the `watchosArm64`, `watchosDeviceArm64` and `watchosSimulatorArm64` targets to all published modules except `-ui`, which stays iOS only (`either` also skips `watchosDeviceArm64`: arrow-core does not publish it). Minimum versions: watchOS 7.0 (9.0 on `watchosDeviceArm64`).

- The two APIs that do not exist on watchOS (`presentCodeRedemptionSheet`, `showStoreMessagesForTypes`) log and no-op there via internal `expect`/`actual` helpers.
- `NSInteger` is 32-bit on arm64_32 (`watchosArm64`), so mapping code handles variable-width types, with new unit tests.
- CI: new self-contained `build-and-test-watchos` job in parallel to the iOS jobs. `update-error-codes` now outputs `errors.apple.kt`, so RevenueCat/purchases-error-codes#24 must merge first.

---

> [!NOTE]
> **Medium Risk**
> New published Kotlin/Native targets and Apple interop mapping changes (including 32-bit NSInteger on watchosArm64) can affect published artifacts and price/error conversions. Existing iOS APIs stay source-compatible; watchOS-only APIs no-op rather than crash.
> 
> **Overview**
> Adds **watchOS** as a published platform (`watchosArm64`, `watchosDeviceArm64`, `watchosSimulatorArm64`) for core, models, mappings, kn-core, and result. `-ui` stays iOS-only; `either` omits `watchosDeviceArm64` because arrow-core does not publish it. Minimum versions: watchOS 7.0 (9.0 on `watchosDeviceArm64`).
> 
> APIs that do not exist on watchOS (`presentCodeRedemptionSheet`, `showInAppMessagesIfNeeded`) are split via `expect`/`actual` so iOS still calls native APIs and watchOS logs and no-ops. Availability messages now mention watchOS where relevant (win-backs, ad tracking, paywall tracking).
> 
> Apple mappings handle **32-bit `NSInteger` on arm64_32**: prices use `longLongValue`, error codes convert via `toLong()`, and `revenueMicros` is clamped with `toNSInteger()`. New appleTest coverage for those conversions.
> 
> CI adds a self-contained `build-and-test-watchos` job (compile, metadata, simulator tests, device link) on non-release and release-branch workflows. Error-code generation now writes `errors.apple.kt`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9124764a9e8bfabc1c3ba57f22d6ad5d5836c69d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,11 +42,6 @@ commands:
           command: sdkmanager "platform-tools"
 
   save-kotlin-native-compiler-to-cache:
-    parameters:
-      # CircleCI caches are write-once per key, so a different dependency set needs its own key.
-      key-suffix:
-        type: string
-        default: ""
     steps:
       - run:
           name: Get the Kotlin version
@@ -56,13 +51,9 @@ commands:
       - save_cache:
           name: Saving Kotlin/Native compiler to cache
           paths: [~/.konan]
-          key: konan<< parameters.key-suffix >>-{{ arch }}-{{ checksum "~/.kotlin_version" }}
+          key: konan-{{ arch }}-{{ checksum "~/.kotlin_version" }}
 
   restore-kotlin-native-compiler-from-cache:
-    parameters:
-      key-suffix:
-        type: string
-        default: ""
     steps:
       - run:
           name: Get the Kotlin version
@@ -71,7 +62,7 @@ commands:
       # containing it instead.
       - restore_cache:
           name: Restoring Kotlin/Native compiler from cache
-          key: konan<< parameters.key-suffix >>-{{ arch }}-{{ checksum "~/.kotlin_version" }}
+          key: konan-{{ arch }}-{{ checksum "~/.kotlin_version" }}
 
   save-gradle-user-home-directory-to-cache:
     steps:
@@ -195,16 +186,13 @@ jobs:
       - revenuecat/install-mise-tools:
           tools: java
       - restore-gradle-user-home-directory-from-cache
-      # Compiles both iOS and watchOS targets: restores both konan caches and saves neither,
-      # so each key keeps its single writer (the iOS jobs / build-and-test-watchos).
       - restore-kotlin-native-compiler-from-cache
-      - restore-kotlin-native-compiler-from-cache:
-          key-suffix: -watchos
       - attach-workspace-at-working-directory
       - run:
           name: Validate binary compatibility
           command: ./gradlew apiCheck
       - save-gradle-user-home-directory-to-cache
+      - save-kotlin-native-compiler-to-cache
 
   build-libraries-android:
     executor: android202409
@@ -263,8 +251,7 @@ jobs:
       - revenuecat/install-mise-tools:
           tools: java
       - restore-gradle-user-home-directory-from-cache
-      - restore-kotlin-native-compiler-from-cache:
-          key-suffix: -watchos
+      - restore-kotlin-native-compiler-from-cache
       - run:
           name: Build libraries and run tests for watchOS
           command: |
@@ -279,7 +266,6 @@ jobs:
               tasks+=(":either:compileKotlinWatchos$arch")
             done
             # The commonized (metadata) compilation only runs at publish time otherwise.
-            # Qualified per module to keep the sample apps out of this job.
             for module in core mappings kn-core models result either; do
               tasks+=(":$module:compileAppleMainKotlinMetadata")
               tasks+=(":$module:watchosSimulatorArm64Test")
@@ -290,8 +276,7 @@ jobs:
             tasks+=(":core:linkDebugTestWatchosDeviceArm64")
             ./gradlew "${tasks[@]}"
       - save-gradle-user-home-directory-to-cache
-      - save-kotlin-native-compiler-to-cache:
-          key-suffix: -watchos
+      - save-kotlin-native-compiler-to-cache
 
   build-sample-android:
     executor: android202409
@@ -456,11 +441,7 @@ jobs:
       - revenuecat/install-gem-mac-dependencies:
           cache-version: v1
       - restore-gradle-user-home-directory-from-cache
-      # Compiles both iOS and watchOS targets: restores both konan caches and saves neither,
-      # so each key keeps its single writer (the iOS jobs / build-and-test-watchos).
       - restore-kotlin-native-compiler-from-cache
-      - restore-kotlin-native-compiler-from-cache:
-          key-suffix: -watchos
       - when:
           condition: << parameters.snapshot >>
           steps:
@@ -474,6 +455,7 @@ jobs:
                 name: "Publish release to Maven Central"
                 command: bundle exec fastlane publish_if_release
       - save-gradle-user-home-directory-to-cache
+      - save-kotlin-native-compiler-to-cache
 
   create-github-release:
     description: "Creates a GitHub release for the current version"
@@ -569,17 +551,14 @@ workflows:
             - revenuecat/install-mise-tools:
                 tools: java,ruby
             - restore-gradle-user-home-directory-from-cache
-            # Compiles both iOS and watchOS targets: restores both konan caches and saves
-            # neither, so each key keeps its single writer.
             - restore-kotlin-native-compiler-from-cache
-            - restore-kotlin-native-compiler-from-cache:
-                key-suffix: -watchos
             - revenuecat/install-gem-mac-dependencies:
                 cache-version: v1
             - run:
                 name: Regenerate API dumps
                 command: ./gradlew apiDump
             - save-gradle-user-home-directory-to-cache
+            - save-kotlin-native-compiler-to-cache
 
   # The release train starts with the release-train workflow, which is automatically 
   # triggered every week, but can also be triggered manually by passing the "action" 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,12 @@ commands:
           command: sdkmanager "platform-tools"
 
   save-kotlin-native-compiler-to-cache:
+    parameters:
+      # Caches are write-once per key, so jobs downloading different K/N dependency sets (e.g.
+      # watchOS) must use their own key to avoid being shadowed by whichever job saves first.
+      key-suffix:
+        type: string
+        default: ""
     steps:
       - run:
           name: Get the Kotlin version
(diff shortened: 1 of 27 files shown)


.circleci/config.yml
```diff
@@ -42,6 +42,12 @@ commands:
           command: sdkmanager "platform-tools"
 
   save-kotlin-native-compiler-to-cache:
+    parameters:
+      # Caches are write-once per key, so jobs downloading different K/N dependency sets (e.g.
+      # watchOS) must use their own key to avoid being shadowed by whichever job saves first.
```
ajpallares: Maybe we can drop this comment? Probably its usage should be clear enough to show the intention of this. NABD though

mappings/src/appleMain/kotlin/com/revenuecat/purchases/kmp/mappings/AdEventTypes.apple.kt
```diff
@@ -64,7 +67,12 @@ public fun AdRevenueData.toIos(): RCAdRevenue {
         placement = placement,
         adUnitId = adUnitId,
         impressionId = impressionId,
-        revenueMicros = revenueMicros,
+        // The native type is NSInteger, which is 32 bits on watchosArm64 (arm64_32). Clamping
+        // to the platform NSInteger range before convert() turns a silent sign-wrap into a
+        // bounded value there; it is a no-op on 64-bit targets.
+        revenueMicros = revenueMicros
+            .coerceIn(NSIntegerMin.toLong(), NSIntegerMax.toLong())
+            .convert(),
```
ajpallares: Good catch here! However, should we log something when the clamp happens? Maybe not very common when working with USD, but it could happen more frequently in other currencies (e.g. JPY)

core/src/appleMain/kotlin/com/revenuecat/purchases/kmp/Purchases.apple.kt
```diff
@@ -707,21 +705,16 @@ public actual class Purchases private constructor(private val iosPurchases: IosP
...
     public actual fun enableAdServicesAttributionTokenCollection() {
-        if (AppleApiAvailability().isEnableAdServicesAttributionTokenCollectionAPIAvailable())
+        if (appleApiAvailability.isEnableAdServicesAttributionTokenCollectionAPIAvailable())
             iosPurchases.attribution().enableAdServicesAttributionTokenCollection()
         else logHandler.d(
             tag = "Purchases",
             msg = "`enableAdServicesAttributionTokenCollection()` is only available on iOS 14.3 " +
-                    "and up."
+                    "and up. It is not available on watchOS."
         )
```
ajpallares, line 722: Should this be split the same way done for `presentCodeRedemptionSheetIfAvailable` (iosMain vs watchosMain), and perhaps mention the watchOS no-op in the common KDoc (same as `presentCodeRedemptionSheet` already does)?

.circleci/config.yml
```diff
@@ -241,6 +254,45 @@ jobs:
...
+            done
+            # The commonized (metadata) compilation only runs at publish time otherwise.
+            # Qualified per module to keep the sample apps out of this job.
+            for module in core mappings kn-core models result either; do
+              tasks+=(":$module:compileAppleMainKotlinMetadata")
+              tasks+=(":$module:watchosSimulatorArm64Test")
+            done
+            # Links an arm64_32 binary: compile alone would miss bad Swift-runtime linker
+            # opts or missing symbols in the static libs, and no other job links for device.
+            tasks+=(":core:linkDebugTestWatchosArm64")
```
ajpallares, line 275: We already link `:core:linkDebugTestWatchosArm64`, which is the 32-bit-pointer watchOS device target (`arm64_32`). That catches Swift linker problems compile would miss. But we don't do the same for `watchosDeviceArm64`, which is the full 64-bit watch target (`arm64-apple-watchos`, Series 9+). It has its own linker flags, and I don't think compiling Kotlin is enough to catch a bad `watchos_device_arm64` flag or a missing symbol in that slice.

Worth adding `:core:linkDebugTestWatchosDeviceArm64` next to the existing link task? We still wouldn't *run* those tests, but I don't think linking needs one.

ajpallares (approved):
Great addition! I think it looks good! I just have some minor comments

.circleci/config.yml
AlvaroBrey: https://github.com/RevenueCat/purchases-kmp/pull/964/commits/e949786d57879b5228e55e61d25fdc0d93883ab5

.circleci/config.yml
AlvaroBrey: Removed

core/src/appleMain/kotlin/com/revenuecat/purchases/kmp/Purchases.apple.kt
AlvaroBrey: Added a KDoc, but split isn't necessary because it exists in watchOS (it just does nothing)

mappings/src/appleMain/kotlin/com/revenuecat/purchases/kmp/mappings/AdEventTypes.apple.kt
AlvaroBrey: Unrelated but in revisiting this I decided to extract the mapper to a [dedicated extension](https://github.com/RevenueCat/purchases-kmp/pull/964/changes/c1359407886cf356906f9a725afeaa7d5479e2ce)  so it's easier to "do it right" in the future.

That being said, the problem here is the fact that the native swift SDK  should've been using `Int64`/`long long` for micros instead of the variable width ones. There is no real fix we can do at the KMP level, only patches (and the fix would be breaking on iOS). 

On top of that, there is no easy access to a logger here, and it wouldn't help anyway since the only action a developer could take would be to complain to us about it 😅 

tldr: this needs to be fixed on `purchases-ios` and it's a breaking change. is it worth it for watchOS support? 🤷

ajpallares: Indeed, it would be indeed a breaking change in purchases-ios. A subtle one, but one nonetheless... I don't think we need to change this just for watchOS (also one that doesn't need fixing in Apple Watch Series 9 and newer)

.circleci/config.yml
```diff
@@ -51,9 +56,13 @@ commands:
       - save_cache:
           name: Saving Kotlin/Native compiler to cache
           paths: [~/.konan]
-          key: konan-{{ arch }}-{{ checksum "~/.kotlin_version" }}
+          key: konan<< parameters.key-suffix >>-{{ arch }}-{{ checksum "~/.kotlin_version" }}
```
JayShortway: Is it not the same konan compiler that compiles for iOS and watchOS? (Note that `{{ arch }}` refers to the host architecture.)

.circleci/config.yml
```diff
@@ -241,6 +253,46 @@ jobs:
...
+              for arch in Arm64 DeviceArm64 SimulatorArm64; do
+                tasks+=(":$module:compileKotlinWatchos$arch")
+              done
+            done
+            # either skips DeviceArm64: arrow-core doesn't publish it.
+            for arch in Arm64 SimulatorArm64; do
+              tasks+=(":either:compileKotlinWatchos$arch")
+            done
+            # The commonized (metadata) compilation only runs at publish time otherwise.
+            # Qualified per module to keep the sample apps out of this job.
```
JayShortway:
```suggestion

```

.circleci/config.yml
```diff
@@ -404,7 +456,11 @@ jobs:
       - revenuecat/install-gem-mac-dependencies:
           cache-version: v1
       - restore-gradle-user-home-directory-from-cache
+      # Compiles both iOS and watchOS targets: restores both konan caches and saves neither,
+      # so each key keeps its single writer (the iOS jobs / build-and-test-watchos).
```
JayShortway:
```suggestion

```

.circleci/config.yml
```diff
@@ -513,14 +569,17 @@ workflows:
             - revenuecat/install-mise-tools:
                 tools: java,ruby
             - restore-gradle-user-home-directory-from-cache
+            # Compiles both iOS and watchOS targets: restores both konan caches and saves
+            # neither, so each key keeps its single writer.
```
JayShortway:
```suggestion

```

mappings/src/appleMain/kotlin/com/revenuecat/purchases/kmp/mappings/Price.apple.kt
```diff
@@ -9,7 +9,7 @@ import com.revenuecat.purchases.kn.core.price
 internal fun RCStoreProduct.toPrice(): Price =
     Price(
         formatted = localizedPriceString(),
-        amountMicros = price().decimalNumberByMultiplyingByPowerOf10(6).longValue,
+        amountMicros = price().decimalNumberByMultiplyingByPowerOf10(6).longLongValue,
```
JayShortway, line 12: Wait, what is `longLongValue`? 😅

JayShortway (commented):
Some questions + some suggestions to remove some LLM comments. Nice set of tests!

mappings/src/appleMain/kotlin/com/revenuecat/purchases/kmp/mappings/Price.apple.kt
AlvaroBrey: [~ long long maaaaan~](https://www.youtube.com/watch?v=6hH4CAz8ET4)

* `amountMicros` in kotlin is a Long
* `longValue` returns `NSInteger` whose width is variable between arches (`Int` in `arm64_32`, `Long` in `arm64`). compilation in `arm64_32` therefore fails because you can't assign an `Int` to a `Long`
* [`longLongValue`](https://developer.apple.com/documentation/foundation/nsstring/longlongvalue) on the other hand always returns a `long long`  (Int64) which is 64 bit regardless of arch and converts to kotlin `Long`

.circleci/config.yml
AlvaroBrey: Yes, actually, it brings all targers. This was just a mistake on my part (conflated build cache vs just the compiler due to reading too quickly). Undoing

AlvaroBrey: done in https://github.com/RevenueCat/purchases-kmp/pull/964/commits/4bb0c476c353d0b4cfbf993872869f6ec63cb44b

mappings/src/appleMain/kotlin/com/revenuecat/purchases/kmp/mappings/Price.apple.kt
JayShortway: Wow TIL, thanks!

JayShortway (approved):

PR #976, opened 2026-08-17
chore(ads): align ad tracking class visibility

peterporfy:
### Description

Closing public construction of AdFormat/AdMediatorName/AdRevenuePrecision to match Android's style. To do this I copied the `InternalRevenueCatApi` annotation from the Kotlin package.

---

> [!NOTE]
> **Medium Risk**
> This is a public ABI break for experimental ad APIs (constructors and `value` access). Low runtime risk, but consumers constructing these types from strings will no longer compile.
> 
> **Overview**
> Makes `AdFormat`, `AdMediatorName`, and `AdRevenuePrecision` closed types: constructors are now `internal`, and `value` is marked with a new **`InternalRevenueCatApi`** opt-in (error-level). Callers should use companion constants instead of constructing from strings.
> 
> Adds `equals`/`hashCode` based on `value`, plus internal `fromString` helpers that map known values to the constants (unknown strings still create a custom instance). Platform mappings opt into the internal API when converting to Android/iOS types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56617d31818dee734519ad0d02fae3bc38dc317b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/InternalRevenueCatApi.kt
+++ b/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/InternalRevenueCatApi.kt
@@ -1,5 +1,11 @@
 package com.revenuecat.purchases.kmp
 
+/**
+ * This annotation marks RevenueCat APIs that are internal and not meant for public consumption.
+ *
+ * APIs marked with this annotation may change frequently and without warning, and are not
+ * covered by semantic versioning guarantees. Do not use in your own code.
+ */
 @Retention(value = AnnotationRetention.BINARY)
 @RequiresOptIn(
     level = RequiresOptIn.Level.ERROR,
--- a/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/InternalRevenueCatApi.kt
+++ b/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/InternalRevenueCatApi.kt
@@ -1,0 +1,16 @@
+package com.revenuecat.purchases.kmp
+
+@Retention(value = AnnotationRetention.BINARY)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.ERROR,
+    message = "This is an internal RevenueCat API that may change frequently and without warning. " +
+        "No compatibility guarantees are provided. It is strongly discouraged to use this API.",
+)
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY,
+    AnnotationTarget.TYPEALIAS,
+    AnnotationTarget.CONSTRUCTOR,
+)
+public annotation class InternalRevenueCatApi
--- a/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
+++ b/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
@@ -29,7 +29,7 @@ public class AdFormat internal constructor(@property:InternalRevenueCatApi publi
         public val NATIVE: AdFormat = AdFormat("native")
         public val APP_OPEN: AdFormat = AdFormat("app_open")
 
-        public fun fromString(value: String): AdFormat {
+        internal fun fromString(value: String): AdFormat {
             return when (value.trim()) {
                 "other" -> OTHER
                 "banner" -> BANNER
--- a/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
+++ b/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
@@ -1,12 +1,13 @@
 package com.revenuecat.purchases.kmp.models
 
 import com.revenuecat.purchases.kmp.ExperimentalRevenueCatApi
+import com.revenuecat.purchases.kmp.InternalRevenueCatApi
 
 /**
  * Common ad format types.
  */
 @ExperimentalRevenueCatApi
-public class AdFormat(public val value: String) {
+public class AdFormat internal constructor(@property:InternalRevenueCatApi public val value: String) {
     public companion object {
         public val OTHER: AdFormat = AdFormat("other")
         public val BANNER: AdFormat = AdFormat("banner")
@@ -15,5 +16,18 @@ public class AdFormat(public val value: String) {
         public val REWARDED_INTERSTITIAL: AdFormat = AdFormat("rewarded_interstitial")
         public val NATIVE: AdFormat = AdFormat("native")
         public val APP_OPEN: AdFormat = AdFormat("app_open")
+
+        public fun fromString(value: String): AdFormat {
+            return when (value.trim()) {
+                "other" -> OTHER
+                "banner" -> BANNER
+                "interstitial" -> INTERSTITIAL
+                "rewarded" -> REWARDED
+                "rewarded_interstitial" -> REWARDED_INTERSTITIAL
+                "native" -> NATIVE
+                "app_open" -> APP_OPEN
+                else -> AdFormat(value)
+            }
+        }
     }
 }
--- a/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
+++ b/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
@@ -8,6 +8,18 @@ import com.revenuecat.purchases.kmp.InternalRevenueCatApi
  */
 @ExperimentalRevenueCatApi
 public class AdFormat internal constructor(@property:InternalRevenueCatApi public val value: String) {
+    @OptIn(InternalRevenueCatApi::class)
+    public override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is AdFormat) return false
+        return value == other.value
+    }
+
+    @OptIn(InternalRevenueCatApi::class)
+    public override fun hashCode(): Int {
+        return value.hashCode()
+    }
+
     public companion object {
         public val OTHER: AdFormat = AdFormat("other")
         public val BANNER: AdFormat = AdFormat("banner")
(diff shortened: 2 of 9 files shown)


peterporfy:
* **#988** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-kmp/988?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#976** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-kmp/976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-kmp/976?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
```diff
@@ -1,12 +1,13 @@
...
 
 import com.revenuecat.purchases.kmp.ExperimentalRevenueCatApi
+import com.revenuecat.purchases.kmp.InternalRevenueCatApi
 
 /**
  * Common ad format types.
  */
 @ExperimentalRevenueCatApi
-public class AdFormat(public val value: String) {
+public class AdFormat internal constructor(@property:InternalRevenueCatApi public val value: String) {
```
tonidero, line 10: Hmm instead of having a `@InternalRevenueCatApi public value` couldn't we just make it `internal`? Do we need to expose it for something? I would think not but not sure... Same applies to the other types in this PR 
```suggestion
public class AdFormat internal constructor(internal val value: String) {
```

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
```diff
@@ -15,5 +16,18 @@ public class AdFormat(public val value: String) {
         public val REWARDED_INTERSTITIAL: AdFormat = AdFormat("rewarded_interstitial")
         public val NATIVE: AdFormat = AdFormat("native")
         public val APP_OPEN: AdFormat = AdFormat("app_open")
+
+        public fun fromString(value: String): AdFormat {
```
tonidero: Maybe this could also be `internal`

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/InternalRevenueCatApi.kt
```diff
@@ -0,0 +1,16 @@
...
+        "No compatibility guarantees are provided. It is strongly discouraged to use this API.",
+)
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY,
+    AnnotationTarget.TYPEALIAS,
+    AnnotationTarget.CONSTRUCTOR,
+)
+public annotation class InternalRevenueCatApi
```
tonidero, line 22: In Android we use
```suggestion
public annotation class InternalRevenueCatAPI
```
But again, not sure if we need this, since I don't think we need to expose these properties anywhere internally right now and we can instead keep things `internal`

tonidero (commented):
TBH, I'm unsure if we need the annotation... Wdyt?

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
AlvaroBrey: Both here and on the other classes:  given this is a plain `class` but it's actually a data model, you likely want to override `equals` and `hashcode` for `fromString` to make sense.

Otherwise `AdFormat.fromString("foo") != AdFormat.fromString("foo")`, which doesn't make sense.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/InternalRevenueCatApi.kt
peterporfy: I used `Api` as to follow existing local convention: https://github.com/RevenueCat/purchases-kmp/blob/ee5e0e490489ab84b188091b40fcdf66d0d6e095/models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/ExperimentalRevenueCatApi.kt?plain=1#L20

But I can change it to `API` for both. If we need it - but I will check the other comment before :) if we keep then what do you think? `API`?

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
peterporfy: If I make it fully internal then the mapping module can't read it here: https://github.com/RevenueCat/purchases-kmp/blob/ee5e0e490489ab84b188091b40fcdf66d0d6e095/mappings/src/androidMain/kotlin/com/revenuecat/purchases/kmp/mappings/AdEventTypes.android.kt?plain=1#L38

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
peterporfy: moved to internal 👍

peterporfy: done!

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
```diff
@@ -15,5 +28,18 @@ public class AdFormat(public val value: String) {
...
+                "other" -> OTHER
+                "banner" -> BANNER
+                "interstitial" -> INTERSTITIAL
+                "rewarded" -> REWARDED
+                "rewarded_interstitial" -> REWARDED_INTERSTITIAL
+                "native" -> NATIVE
+                "app_open" -> APP_OPEN
+                else -> AdFormat(value)
+            }
+        }
```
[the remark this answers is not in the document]
peterporfy: There is no need to build custom AdFormats etc.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
```diff
@@ -15,5 +28,18 @@ public class AdFormat(public val value: String) {
...
+            return when (value.trim()) {
+                "other" -> OTHER
+                "banner" -> BANNER
+                "interstitial" -> INTERSTITIAL
+                "rewarded" -> REWARDED
+                "rewarded_interstitial" -> REWARDED_INTERSTITIAL
+                "native" -> NATIVE
+                "app_open" -> APP_OPEN
+                else -> AdFormat(value)
+            }
```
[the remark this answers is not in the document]
peterporfy: There is no need to build custom values - therefore " FOO " should not happen in reality.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/InternalRevenueCatApi.kt
JayShortway: My vote would be `Api`, as that's the Kotlin convention, but it's not a hill I'd die on. 

From [here](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names):
>For acronyms longer than two letters, capitalize only the first letter. For example, `XmlFormatter` or `HttpInputStream`.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
JayShortway: Yea this is indeed why I suggested the annotation.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
JayShortway: We use [`@Poko`](https://github.com/drewhamilton/Poko) in purchases-android for plain classes that are actually data models. Could be nice to use that here too? That avoids the need for manual `equals` and `hashCode` functions. (Not a blocker of course.)

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/InternalRevenueCatApi.kt
tonidero: Ahh ok, happy to keep it like this then!

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
tonidero: Ohhh ok.... Yeah I missed that. Sounds good then!

AlvaroBrey (approved):

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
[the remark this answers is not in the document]
peterporfy: They serve different purposes so casing is good. They are called internally.

models/src/commonMain/kotlin/com/revenuecat/purchases/kmp/models/AdFormat.kt
peterporfy: @JayShortway https://github.com/RevenueCat/purchases-kmp/pull/988

JayShortway: Nice, approved!

PR #980, opened 2026-08-19
Fix iOS crash in setAttributes with null values

### LINKED ISSUE #979: setAttributes with a null value crashes on iOS (NSNull cannot be cast to NSString)
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-kmp/issues)

**Describe the bug**

`Purchases.setAttributes()` documents that a `null` value deletes an attribute. On Android that works. On iOS it crashes the app with `Could not cast value of type 'NSNull' to 'NSString'`, because the iOS actual forwards Kotlin `null` values into a Swift API whose signature is `[String: String]` (non-optional).

Since `setAttributes` is the only map-based way to clear an attribute, following the documented contract is currently fatal on iOS.

1. Environment
    1. Platform: Kotlin Multiplatform, iOS target (Android unaffected)
    2. SDK version: `purchases-kmp` 3.4.0 — the code on `main` is unchanged at the time of writing, see links below
    3. OS version: iOS 18 (not version specific)
    4. IDE (e.g. Android Studio, Xcode, Fleet): Android Studio + Xcode
    5. IDE version: current stable
        - [x] Device
        - [x] Emulator/simulator
    7. Environment:
        - [ ] TestFlight
        - [ ] Production
    8. How widespread is the issue. Percentage of devices affected: 100% of iOS devices, on any `setAttributes` call where at least one value is `null`. Deterministic, not device or OS specific.

The failure is an immediate uncaught Objective-C cast exception, so it happens before SDK logging is informative:

```
Could not cast value of type 'NSNull' to 'NSString'
```

In our app the attribute sync runs during startup, so the result was a crash on launch.

```kotlin
// iOS target
Purchases.sharedInstance.setAttributes(mapOf("\$email" to null))
```

- **Expected:** the `$email` attribute is deleted, per the KDoc on the common declaration: *"Map of attributes by key. Set the value as null to delete an attribute."*
- **Actual (iOS):** the app crashes with `Could not cast value of type 'NSNull' to 'NSString'`.
- **Actual (Android):** works as documented, attribute is cleared.

4. **Other information**

The cause is visible by comparing the two actuals.

Common declaration — [`Purchases.kt`](https://github.com/RevenueCat/purchases-kmp/blob/main/core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt):

```kotlin
/**
 * @param attributes Map of attributes by key. Set the value as null to delete an attribute.
 */
public fun setAttributes(attributes: Map<String, String?>)
```

Android actual — [`Purchases.android.kt`](https://github.com/RevenueCat/purchases-kmp/blob/main/core/src/androidMain/kotlin/com/revenuecat/purchases/kmp/Purchases.android.kt) (honors the contract):

```kotlin
public actual fun setAttributes(attributes: Map<String, String?>): Unit =
    androidPurchases.setAttributes(attributes)
```

iOS actual — [`Purchases.ios.kt`](https://github.com/RevenueCat/purchases-kmp/blob/main/core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/Purchases.ios.kt) (only remaps keys, so `null` values pass through and become `NSNull`):

```kotlin
public actual fun setAttributes(attributes: Map<String, String?>): Unit =
    iosPurchases.setAttributes(attributes.mapKeys { (key, _) -> key })
```

The receiving Swift signature cannot hold `NSNull`, from [`revenuecat-api-ios.swiftinterface`](https://github.com/RevenueCat/purchases-ios/blob/main/api/revenuecat-api-ios.swiftinterface):

```swift
@objc final public func setAttributes(_ attributes: [Swift.String : Swift.String])
```

Worth noting that the typed setters in the same file do accept `null` on iOS, because their Swift counterparts are optional:

```kotlin
public actual fun setEmail(email: String?): Unit = iosPurchases.setEmail(email)
```

So `null`-means-delete works through `setEmail` but crashes through `setAttributes`, on the same platform, in the same class.

Because Kotlin/Native interop treats the Swift map as a platform type, this compiles cleanly and only fails at runtime on device.

**Suggested fix**

Substitute empty string for `null` in the iOS actual:

```kotlin
public actual fun setAttributes(attributes: Map<String, String?>): Unit =
    iosPurchases.setAttributes(attributes.mapValues { (_, value) -> value ?: "" })
```

RevenueCat's own documentation for the typed setters states *"Null or empty will delete the subscriber attribute"*, so empty string preserves the documented delete semantics on both platforms. We have been shipping exactly this substitution client-side for about three months across Android and iOS, and attribute clearing behaves correctly on the dashboard.

This is contained entirely in the KMP bridge. `purchases-android` and `purchases-ios` both behave correctly on their own, so no follow-up issues appear to be needed there.

I am happy to open a PR with the one-line change if that is useful.

**Additional context**

We hit this while syncing subscriber attributes (`$email`, `$displayName`, `firebase_uid`, and similar) after sign-in. Sending explicit `null` for missing fields is necessary in our case because omitting a key leaves the previous user's value in place after an account switch, which is precisely what the documented `null` behavior is for.

joshdholtz:
### Checklist

- [x] If applicable, create follow-up issues for `purchases-android`, `purchases-ios` and other hybrids

### Motivation

Resolves #979

When calling `Purchases.setAttributes()` with a `null` value on iOS, the app crashes with the error "Could not cast value of type 'NSNull' to 'NSString'". The documented behavior states that setting a value to `null` should delete that subscriber attribute, but on iOS this resulted in a fatal crash, making it impossible to use the documented API.

### Description

The iOS actual now converts `null` attribute values to empty strings before calling the native SDK, which takes `[String: String]` (non-optional). Empty string is the iOS SDK's documented delete representation. Android is unchanged and still sends JSON null.

```kotlin
public actual fun setAttributes(attributes: Map<String, String?>): Unit =
    iosPurchases.setAttributes(attributes.toIosSubscriberAttributes())
```

`toIosSubscriberAttributes()` is the production mapping. Unit tests call that helper rather than re-implementing `mapValues` inline, so a revert of the conversion fails CI. Calling `setAttributes` itself from iOS unit tests is not viable: it requires a configured native SDK instance and previously segfaulted in the simulator. The sample app's subscriber attributes screen already reproduces the original crash with no extra test code.

KDoc on the common `setAttributes` now notes the platform payload difference: iOS sends `""`, Android sends JSON `null`.

No follow-up issues for `purchases-android` or `purchases-ios`. Both native SDKs already handle their own delete representations correctly; only the KMP iOS bridge was missing the conversion.

---

> [!NOTE]
> **Low Risk**
> Small iOS-only mapping change for subscriber attributes; behavior matches the native SDK’s documented delete representation.
> 
> **Overview**
> Stops iOS from crashing when `Purchases.setAttributes` is given `null` values (Kotlin `null` was becoming `NSNull` and failing an `NSString` cast).
> 
> The Apple actual now maps nulls to `""` via `toIosSubscriberAttributes()`, which is the iOS SDK’s documented way to delete an attribute. Android is unchanged. Unit tests cover the mapping helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34b102d6ff87d89d020b0c7dc185d32d76051274. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


core/src/commonMain/kotlin/com/revenuecat/purchases/kmp/Purchases.kt
```diff
@@ -620,6 +620,8 @@ public expect class Purchases {
      * restrictions refer to our [guide](https://docs.revenuecat.com/docs/subscriber-attributes).
      *
      * @param attributes Map of attributes by key. Set the value as null to delete an attribute.
+     * The delete payload differs by platform: iOS sends an empty string (the native SDK's
+     * documented delete representation), while Android sends JSON null.
```
JayShortway: This comment doesn't seem intended for app developers?

core/src/iosMain/kotlin/com/revenuecat/purchases/kmp/SubscriberAttributes.ios.kt
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Maps KMP subscriber attributes onto the iOS SDK's NSDictionary-backed parameter type.
+ *
+ * Kotlin `null` values would otherwise become `NSNull` and crash when the native API tries to
+ * cast them to `NSString`. Empty string is the iOS SDK's documented representation for deleting
+ * an attribute.
+ *
+ * The return type is declared as `Map<Any?, *>` because that is the ObjC interop parameter type
+ * for `RCPurchases.setAttributes`, and [Map] is invariant, so `Map<String, String>` is not a
+ * subtype of it.
```
JayShortway:
```suggestion

```

core/src/iosTest/kotlin/com/revenuecat/purchases/kmp/PurchasesSetAttributesTest.kt
```diff
@@ -0,0 +1,79 @@
...
+    fun `toIosSubscriberAttributes leaves an empty map empty`() {
+        assertEquals(
+            emptyMap<String, String>(),
+            emptyMap<String, String?>().toIosSubscriberAttributes().stringValues()
+        )
+    }
+}
+
+private fun Map<Any?, *>.stringValues(): Map<String, String> =
+    entries.associate { (key, value) -> key as String to value as String }
```
JayShortway: Why do we need this? Why don't we assert the actual values?

JayShortway (commented):
Thanks for fixing! Some minor comments and questions.

ajpallares (approved):
Makes sense to me!

JayShortway (approved):
Thanks for fixing!

PR #147, opened 2024-08-14
Refactor StoreProduct and its dependencies

vegaro:
Migrate away from typealiases. `StoreProduct` and dependencies

PR #981, opened 2026-08-20
renovate: force a release when native SDKs are bumped

ajpallares:
### Motivation
KMP's native SDK bumps come from Renovate (purchases-android via gradle, purchases-ios via git submodule) and carry only the inherited `pr:dependencies` label. Since RevenueCat/fastlane-plugin-revenuecat_internal#147 makes `pr:dependencies` release-neutral, those bumps stopped triggering a KMP release. This restores that behavior.

### Summary
- Add `pr:force_patch` / `pr:force_minor` (by Renovate update type) to the `purchases-android` and `purchases-ios` update PRs, keeping `pr:dependencies` for Danger and changelog categorization.
- Major native bumps get no force label, so they no longer auto-release and must be handled intentionally.

### Notes
- Companion to plugin PR #147 and RevenueCat/purchases-hybrid-common#1781 (PHC's equivalent fix, done in its Fastfile bump lane).

---

> [!NOTE]
> **Low Risk**
> Configuration-only change to Renovate labels; no runtime or SDK code paths are modified.
> 
> **Overview**
> Adds Renovate `packageRules` so **patch** and **minor** bumps of `purchases-android` (Gradle) and `purchases-ios` (git submodule) get **`pr:force_patch`** or **`pr:force_minor`** on the PR, alongside the existing **`pr:dependencies`** label.
> 
> That restores automatic KMP releases when native SDKs are bumped via Renovate, after release tooling treated dependency-only PRs as release-neutral. **Major** native updates still get no force label, so they do not auto-release.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7d131e367f067a9ea23853cb3be41c4b84d384b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -18,6 +18,28 @@
       "matchManagers": ["gradle"],
       "enabled": true,
       "groupName": "purchases-android"
+    },
+    {
+      "description": "Force a patch release when the native SDKs get a patch update",
+      "matchManagers": ["gradle", "git-submodules"],
+      "matchPackageNames": [
+        "com.revenuecat.purchases:purchases",
+        "com.revenuecat.purchases:purchases-ui",
+        "upstream/purchases-ios"
+      ],
+      "matchUpdateTypes": ["patch"],
+      "addLabels": ["pr:force_patch"]
+    },
+    {
+      "description": "Force a minor release when the native SDKs get a minor update",
+      "matchManagers": ["gradle", "git-submodules"],
+      "matchPackageNames": [
+        "com.revenuecat.purchases:purchases",
+        "com.revenuecat.purchases:purchases-ui",
+        "upstream/purchases-ios"
+      ],
+      "matchUpdateTypes": ["minor"],
+      "addLabels": ["pr:force_minor"]
     }
   ]
 }
--- a/renovate.json
+++ b/renovate.json
@@ -20,24 +20,36 @@
       "groupName": "purchases-android"
     },
     {
-      "description": "Force a patch release when the native SDKs get a patch update",
-      "matchManagers": ["gradle", "git-submodules"],
+      "description": "Force a patch release when purchases-android gets a patch update",
+      "matchManagers": ["gradle"],
       "matchPackageNames": [
         "com.revenuecat.purchases:purchases",
-        "com.revenuecat.purchases:purchases-ui",
-        "upstream/purchases-ios"
+        "com.revenuecat.purchases:purchases-ui"
       ],
       "matchUpdateTypes": ["patch"],
       "addLabels": ["pr:force_patch"]
     },
     {
-      "description": "Force a minor release when the native SDKs get a minor update",
-      "matchManagers": ["gradle", "git-submodules"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -18,6 +18,28 @@
       "matchManagers": ["gradle"],
       "enabled": true,
       "groupName": "purchases-android"
+    },
+    {
+      "description": "Force a patch release when the native SDKs get a patch update",
```
tonidero: Just to check, do we need the above 2 package rules? Seems like we would be covered with these 2? Though maybe not on majors... not sure 😅

tonidero (approved):
Good catch! Just a question

renovate.json
ajpallares: Hmm, I think we still need them, but I'm not very experienced with this so I could be wrong:

The shared preset starts with `{"matchPackageNames": ["*"], "enabled": false}`, and presets are resolved first. So these two rules are what actually re-enable the dependencies. The new rules only `addLabels`, they never set `enabled`.

```
{
  "matchManagers": ["git-submodules"],
  "enabled": true
}
```

enables the purchases-ios submodule (it's our only submodule).

```
{
  "matchPackageNames": [
    "com.revenuecat.purchases:purchases",
    "com.revenuecat.purchases:purchases-ui"
  ],
  "matchManagers": ["gradle"],
  "enabled": true,
  "groupName": "purchases-android"
}
```

enables the two purchases-android artifacts and groups them into a single PR.

Also, even if we moved `enabled: true` into the new rules, they only match patch/minor update types, so majors would fall back to the preset's disable and Renovate would never open a major native bump PR. With this, it does, just without a force label so it doesn't auto-release.

tonidero: Right, makes sense! Happy to leave as it is 👍

AlvaroBrey (approved):

renovate.json
```diff
@@ -18,6 +18,28 @@
...
+      "matchManagers": ["gradle", "git-submodules"],
+      "matchPackageNames": [
+        "com.revenuecat.purchases:purchases",
+        "com.revenuecat.purchases:purchases-ui",
+        "upstream/purchases-ios"
+      ],
```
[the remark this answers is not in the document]
ajpallares: Good catch. Renovate's `git-submodules` manager sets `depName` to the submodule path (`upstream/purchases-ios`) but `packageName` to the submodule's repo URL, and `matchPackageNames` only matches `packageName`, so the iOS submodule never matched these rules.

Fixed by splitting each force rule in two, since matchers inside a single rule are ANDed and can't express "gradle coordinate OR submodule dep": the gradle rules keep `matchPackageNames` and the new iOS rules use `matchDepNames`.

@tonidero and/or @AlvaroBrey you may want to give it another quick look. Sorry for the noise 🙏

AlvaroBrey (approved):

PR #1015, opened 2026-09-10
feat(paywalls): Add onInteraction callback to PaywallListener

AlvaroBrey:
- Adds `PaywallListener.onInteraction(event: PaywallInteractionEvent)`, a default no-op, fired for every `paywall_component_interacted` the native paywall tracks so apps can route them to their analytics tools.
- `PaywallInteractionEvent` mirrors the purchases-android class: `rawProperties` is the snake_case map, `Keys` are typed accessors read via `getProperty(Keys.X)`, `ComponentTypes` lists the known `component_type` values.
- Wires a `commonTest` source set for `revenuecatui` (the module had none) and regenerates the JVM and klib API dumps.

Part of [SDK-4457](https://linear.app/revenuecat/issue/SDK-4457).

### Checklist

  hybrids

<details><summary>Agent description</summary>

### Motivation

The native SDKs expose paywall interactions to developers (purchases-android 10.21.0, purchases-ios 5.89.0). KMP wraps both natives directly rather than through PHC, so it needs its own listener method and event type to offer the same callback.

### Description

- `PaywallListener.kt`: `onInteraction(event: PaywallInteractionEvent) {}`, last in the interface, same shape as `onUrlOpened`.
- `PaywallInteractionEvent.kt` (commonMain): `rawProperties: Map<String, Any>`, `getProperty(key: Key<T>): T?`, `Keys` (27 typed keys, `KClass`-backed since `Class<T>` is JVM-only), `ComponentTypes` (8 string constants). `equals`/`hashCode`/`toString` delegate to the map, as on Android.
- Android actual (`PaywallOptionsKtx.kt`): wraps `event.rawProperties` without copying.
- iOS actual (`PaywallOptionsKtx.kt`): implements `paywallViewController(_:didTrackInteraction:)` on `IosPaywallDelegate` and converts the `NSDictionary`. `NSNumber` needs `objCType` to tell Swift `Bool` (`"c"`) from integers; the contract has no floating-point keys, so every other `NSNumber` becomes `Long`.
- `getProperty` widens `Number` to `Int`/`Long` per key type because of that iOS behavior; `Key` types are otherwise checked with `safeCast`, so a wrong type reads as null rather than throwing.
- `revenuecatui/build.gradle.kts`: `commonTest` and `androidUnitTest` dependencies, copied from `models`.
- `apiTester`: `PaywallListenerAPI` exercises the override, `getProperty` with each key type, and `ComponentTypes`.

### Regression gates

- `PaywallInteractionEventTest.integer keys accept any numeric width`: fails if the widening in `getProperty` is removed (iOS would then return null for every `Int` key).
- The rest of `PaywallInteractionEventTest` covers typed reads, absent keys and mismatched types.

Not visible in the diff: the iOS `NSNumber` branch has no unit test, since `revenuecatui` has no Apple test source set; the api tester and the sample app are the checks there.

</details>

---

> [!NOTE]
> **Low Risk**
> Additive public API with a default no-op callback; purchase flows are unchanged aside from new listener forwarding on Android and iOS.
> 
> **Overview**
> Adds **`PaywallListener.onInteraction`** (default no-op) so KMP apps can receive native paywall control interactions (tabs, packages, purchase button, etc.) and forward them to analytics, matching purchases-android/iOS.
> 
> Introduces **`PaywallInteractionEvent`** with a snake_case **`rawProperties`** map, typed **`getProperty(Keys.*)`** accessors, and **`ComponentTypes`** constants. Android forwards the native event’s properties; iOS handles **`didTrackInteraction`** and normalizes **`NSNumber`** values (bools vs integers) before building the KMP event.
> 
> Also adds **`commonTest`** coverage for **`PaywallInteractionEvent`**, updates the API tester and JVM/klib API dumps, and wires **`revenuecatui`** unit test dependencies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb546fb7224606a856c44408662badec40703ab0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlvaroBrey:
@cursor review

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallOptionsKtx.kt
```diff
@@ -105,3 +117,13 @@ internal class IosPaywallDelegate(
         onHeightChange(height!!)
     }
 }
+
+private fun Map<Any?, *>.toKotlinValues(): Map<String, Any> =
+    entries.associate { (key, value) ->
+        key as String to if (value is NSNumber) value.toKotlinValue() else value as Any
```
tonidero: Hmm should we cast these safely and log an error if something fails? Shouldn't happen... but you never know 😅

tonidero (approved):

revenuecatui/src/iosMain/kotlin/com/revenuecat/purchases/kmp/ui/revenuecatui/PaywallOptionsKtx.kt
AlvaroBrey: Added some defensive null and type checks [here](https://github.com/RevenueCat/purchases-kmp/pull/1015/commits/1cf5d08f0f9769a83072e396ee16708b7ba5c059) although they are theoretically not reachable, since the dict is not nullable in Swift/ObjC anyway, and the real types should always be String or NSNumber.

However I'm not sure what you mean by "if something fails", so let me know if you meant something else pls

tonidero (approved):
Yup, I think changes are good. Thank you!!

PR #1016, opened 2026-09-11
Update purchases-ios to 5.89.0 + ads compatibility

AlvaroBrey:
- Moves `upstream/purchases-ios` to `5.89.0`.
- `purchases-ios` `5.89.0` dropped the Objective-C names of the ads event types and `AdTracker`, and marked ads and reward verification `@available(iOS 15.0, watchOS 8.0)`. Kotlin/Native only sees the Objective-C header, so `AdEventTypes.apple.kt` and `Purchases.adTracker()` no longer resolve.
- Routes ad tracking through a new `@objc` `AdTracking` facade in `kn-core`'s `AdditionalSwift`, the same shape as the existing `RewardVerification` wrapper; `AdTracker.apple.kt` calls it with plain strings and numbers.
- `generateRewardVerificationToken` now throws `IllegalStateException` on iOS below 15 and watchOS below 8, and `pollRewardVerification` completes with `failed = true` there. Both used to compile against a non-gated API; on `5.88.0` they ran on iOS 13 and 14.
- Removes `AdEventTypes.apple.kt`, its `appleTest`, and the now unused `Long.toNSInteger()`.

### Checklist

  hybrids

<details><summary>Agent description</summary>

### Motivation

`purchases-ios` [#7490](https://github.com/RevenueCat/purchases-ios/pull/7490) (in `5.89.0`) removed `@objc(RCAd*)` from `MediatorName`, `AdFormat`, `AdDisplayed`, `AdOpened`, `AdLoaded`, `AdFailedToLoad`, `AdRevenue` and `AdRevenue.Precision`, left `Purchases.adTracker` and `AdTracker` Swift-only, and gated them plus `generateRewardVerificationToken` / `pollRewardVerification` behind iOS 15 / watchOS 8. The Renovate bump to `5.89.0` and [#1015](https://github.com/RevenueCat/purchases-kmp/pull/1015) (which needs `5.89.0` for the paywall `onInteraction` delegate) both fail to compile `core` for Apple targets without this.

### Description

- `kn-core/src/swift/Sources/AdditionalSwift/AdTracking.swift`: `@objc` static `trackAdDisplayed`, `trackAdOpened`, `trackAdLoaded`, `trackAdRevenue`, `trackAdFailedToLoad` taking the KMP model fields as strings and numbers and building the `RevenueCat` structs. The class carries the `@available` annotation; the OS check stays on the Kotlin side through `AppleApiAvailability.isAdTrackingAPIAvailable()`, like the win-back APIs, since cinterop drops availability.
- `RewardVerification.swift`: same class-level `@available` so it compiles against the gated upstream methods; no behavior change inside.
- `AdTracker.apple.kt`: constructor takes nothing, each method checks availability, logs, and calls the facade. `mediatorErrorCode` crosses as `NSNumber?`.
- `Purchases.apple.kt`: `adTracker` no longer reads `iosPurchases.adTracker()`; the token call checks availability and throws; the poll call logs and completes with a failed `RewardVerificationResult` without calling into Swift.
- `Purchases.kt` (common): `@throws` added to the token KDoc.
- `revenueMicros` crosses as `Int64` and is clamped to `NSInteger` in Swift with `Int(clamping:)`. Not visible in the diff: the deleted `AdEventTypesTest` guarded a hand-rolled Kotlin clamp that no longer exists; the `Int64` parameter type is the gate now, since a `Long` would not compile into an `NSInteger` parameter on `watchosArm64`.

### Regression gates

- `core` Apple compile is the gate for the cinterop surface: it fails on `5.89.0` without the facade.
- No unit test covers the facade; `kn-core` has no Swift test target and `mappings/appleTest` cannot reach it.

**Rejected:**
- Mirroring the Swift types as `@objc` classes (like `VerifiedReward`): the ad events are inputs only, so five mirror classes would be `AdEventTypes.apple.kt` rewritten in Swift with the same field-by-field maintenance.
- Reusing `RewardedAdTrackingMetadata` as the parameter of the displayed / opened / loaded methods: saves parameter lists but names a reward-verification type in every ad event call.
- Guarding with `#available` inside each Swift method and returning a synthetic result: duplicates the Kotlin check that already owns the log message, and needed a hand-built `RewardVerificationResult`.
- Returning an empty `RewardVerificationToken` below iOS 15 instead of throwing: fails silently on the ad network's server later.

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes how ads and reward verification cross the KMP/Apple boundary and tightens runtime behavior on iOS 14 and below (token now throws; poll fails explicitly), though guards mirror upstream availability.
> 
> **Overview**
> Bumps the **purchases-ios** submodule to **5.89.0** and restores Apple KMP compilation after upstream made ad tracking and reward verification Swift-only and **iOS 15+ / watchOS 8+** gated (no longer visible to Kotlin/Native cinterop).
> 
> **Ad tracking** no longer goes through `RCAdTracker` and `AdEventTypes.apple.kt` mappings. A new **`@objc` `AdTracking`** Swift facade (same pattern as `RewardVerification`) forwards plain strings/numbers to `Purchases.shared.adTracker`; **`AdTracker.apple.kt`** calls it after `AppleApiAvailability` checks. The Kotlin **`AdEventTypes`** mappers, **`Long.toNSInteger()`**, and their **apple tests** are removed.
> 
> **Reward verification** on Apple now enforces the same OS floor: **`generateRewardVerificationToken`** throws **`IllegalStateException`** when unavailable (documented in common API); **`pollRewardVerification`** logs and completes with **`failed = true`** instead of calling Swift. **`RewardVerification.swift`** adds class-level **`@available`** for the gated upstream API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9f1df5d83dc2e9ebac94623827c9319ca9a4f4f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kn-core/src/swift/Sources/AdditionalSwift/AdTracking.swift
```diff
@@ -0,0 +1,108 @@
+import Foundation
+import RevenueCat
+
+/// Bridges `Purchases.shared.adTracker` into `@objc`-visible entry points, since Kotlin/Native cinterop only
+/// sees declarations reachable from the generated Objective-C header and the ad tracking API is Swift-only.
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+@objc
+public class AdTracking: NSObject {
```
tonidero, line 8: I believe we haven't added swift code to this version of the KMP library before right? I think it should work... But probably good to verify at runtime (if you haven't already 😅)

kn-core/src/swift/Sources/AdditionalSwift/AdTracking.swift
```diff
@@ -0,0 +1,108 @@
...
+    @objc
+    public static func trackAdDisplayed(
+        networkName: String?,
+        mediatorName: String,
+        adFormat: String,
+        placement: String?,
+        adUnitId: String,
+        impressionId: String
+    ) {
+        Purchases.shared.adTracker.trackAdDisplayed(.init(
```
tonidero, line 19: Should we add some check on whether the SDK is configured? Maybe not needed since it seems we weren't before...

tonidero (approved):
I think it should work!

kn-core/src/swift/Sources/AdditionalSwift/AdTracking.swift
peterporfy: If we would check that I think we should in the `purchases-ios` - which currently only checks when admob adapter is used. I think that is might need to change (but not here)

peterporfy:
Thanks for this patch and sorry for the trouble!

peterporfy (approved):
Tested it locally too 👍

kn-core/src/swift/Sources/AdditionalSwift/AdTracking.swift
AlvaroBrey: We have, here in the same `AdditionalSwift` we have both `AppleApiAvailability` and `RewardVerification` (though that one is recent too).

tonidero: Ah right good point!

kn-core/src/swift/Sources/AdditionalSwift/AdTracking.swift
peterporfy: I checked this and it will crash on `Purchases.shared` access in case the sdk is not configured - as a generic not configured error: https://github.com/RevenueCat/purchases-ios/blob/3f8eb5c64ba5a451be2f2a1c2805b761fb952c7f/Sources/Purchasing/Purchases/Purchases.swift#L73

I think that is good - and as mentioned `isConfigured` can be used by someone if needed. What do you think?

